PR #3, opened 2021-07-07
added spring-boot-starter-web dependency

Darina007:


pom.xml
```diff
@@ -27,6 +27,12 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-web</artifactId>
+			<version>2.5.2</version>
```
AntonMaksimenkau: Can we skip the version here? We have a parent "spring-boot-starter-parent" with 2.5.2 version. All the spring versions should use it by default. But having specified versions in the pom file will increase update complexity

maxim12321 (dismissed):

AntonMaksimenkau (approved):

maxim12321 (approved):

PR #8, opened 2021-07-09
added crud for questions

Darina007:


src/main/java/com/team4/testingsystem/controllers/QuestionController.java
```diff
@@ -0,0 +1,25 @@
...
+public class QuestionController {
+    private final QuestionService questionService;
+
+    @Autowired
+    public QuestionController(QuestionService questionService) {
+        this.questionService = questionService;
+    }
+
+    @GetMapping("/{id}")
+    public Question getLevel(@PathVariable("id") Long id) {
```
AntonMaksimenkau: Does the method really return level?

src/main/java/com/team4/testingsystem/repositories/LevelRepository.java
```diff
@@ -0,0 +1,9 @@
+package com.team4.testingsystem.repositories;
+
+import com.team4.testingsystem.entities.Level;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface LevelRepository extends CrudRepository<Level, Integer> {
```
AntonMaksimenkau: As I seem Level's PK (id) has a Long type. Why did we use Integer here?

src/main/java/com/team4/testingsystem/repositories/ModuleRepository.java
```diff
@@ -0,0 +1,9 @@
+package com.team4.testingsystem.repositories;
+
+import com.team4.testingsystem.entities.Module;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface ModuleRepository extends CrudRepository<Module, Integer> {
```
AntonMaksimenkau: Same here

src/main/java/com/team4/testingsystem/repositories/QuestionRepository.java
```diff
@@ -0,0 +1,12 @@
+package com.team4.testingsystem.repositories;
+
+import com.team4.testingsystem.entities.Question;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+
+import java.util.Optional;
+
+@Repository
+public interface QuestionRepository extends CrudRepository<Question, Integer> {
```
AntonMaksimenkau: And here

src/main/java/com/team4/testingsystem/services/impl/QuestionServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+    private QuestionRepository questionRepository;
+
+    @Autowired
+    QuestionServiceImpl(QuestionRepository questionRepository) {
+        this.questionRepository = questionRepository;
+    }
+
+    @Override
+    public Question getQuestionById(Long id) {
+        return questionRepository.getQuestionById(id).orElse(null);
```
AntonMaksimenkau: Please, throw errors instead of returning nulls

src/main/java/com/team4/testingsystem/controllers/QuestionController.java
```diff
@@ -0,0 +1,25 @@
...
+        return questionService.getQuestionById(id);
+    }
```
AntonMaksimenkau, line 50: I think we need an ability to create, update or delete question as well

AntonMaksimenkau (changes requested):
Please, have a look at the comments

src/main/java/com/team4/testingsystem/controllers/QuestionController.java
Darina007: my fault 😰
I'll fix it now

src/main/java/com/team4/testingsystem/controllers/QuestionController.java
Darina007: do we need to delete questions? they should be used in tests and then to display hr. If we want to delete, then we need to check if this question is used in tests?

AntonMaksimenkau: Yep, you are right. Alternatively, we can block the question. Anyway, it will be a separate task.
Still, we need a way to create and update the questions

src/main/java/com/team4/testingsystem/entities/Question.java
```diff
@@ -0,0 +1,83 @@
...
+    private Long id;
+    @Column(name = "question_body")
+    private String questionBody;
+    private int points;
+    @Column(name = "is_avaliable")
+    private int isAvailable;
+
+    @OneToOne
+    @JoinColumn(name = "id", insertable = false, updatable = false)
+    private User creatorUserId;
```
maxim12321: Is Question-CreatorUser relation really OneToOne?
It seems that one user can create more than one question

src/main/java/com/team4/testingsystem/services/impl/QuestionServiceImpl.java
```diff
@@ -0,0 +1,23 @@
...
+import com.team4.testingsystem.entities.Question;
+import com.team4.testingsystem.repositories.QuestionRepository;
+import com.team4.testingsystem.services.QuestionService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class QuestionServiceImpl implements QuestionService {
+
+    private QuestionRepository questionRepository;
```
maxim12321: Maybe make this field final?

src/main/java/com/team4/testingsystem/entities/Question.java
Darina007: Yes, i think you are right
Thanks)

src/main/java/com/team4/testingsystem/services/impl/QuestionServiceImpl.java
Darina007: Maybe 😄

src/main/java/com/team4/testingsystem/entities/Level.java
```diff
@@ -0,0 +1,28 @@
...
+
+import javax.persistence.*;
+
+@Entity
+public class Level {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @Column(name = "level_name")
+    private String levelName;
```
AntonMaksimenkau, line 11: It is a level class. The level is already in context. We don't need to add a level prefix for the fields

AntonMaksimenkau: We are in a level class. It already has level context. You don't need to repeat it with a level prefix

src/main/java/com/team4/testingsystem/entities/Module.java
```diff
@@ -0,0 +1,28 @@
...
+
+import javax.persistence.*;
+
+@Entity
+public class Module {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @Column(name = "module_name")
+    private String moduleName;
```
AntonMaksimenkau: same here

src/main/java/com/team4/testingsystem/entities/Question.java
```diff
@@ -0,0 +1,83 @@
...
+
+import javax.persistence.*;
+
+@Entity
+public class Question {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+    @Column(name = "question_body")
+    private String questionBody;
```
AntonMaksimenkau: same here

src/main/java/com/team4/testingsystem/entities/Question.java
AntonMaksimenkau: The field is storing user, that created a question. Let`s call it something like "creator".
User - is in context.
Id - is about FK in a database. It is a separate layer and we don't need to propagate it to a system.

src/main/java/com/team4/testingsystem/entities/Question.java
```diff
@@ -0,0 +1,83 @@
...
+    private int points;
```
AntonMaksimenkau: Do we need to store the points in a DB?

src/main/java/com/team4/testingsystem/entities/Question.java
```diff
@@ -0,0 +1,83 @@
...
+    @Column(name = "is_avaliable")
+    private int isAvailable;
```
AntonMaksimenkau: Why do "isAvailable" is an integer? Do we really have lots of reasons why do the question is unavailable and want to store it in a such way?
BTW. Field "isAvailable" with getters and setters like getIsAvailable/setIsAvailable seems weird. How do you think, can we use Boolean available with isAvailable(getAvailable) and setAvailable methods?

src/main/java/com/team4/testingsystem/entities/Question.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @OneToOne
+    @JoinColumn(name = "id", insertable = false, updatable = false)
+    private User creatorUserId;
+
+    @ManyToOne
+    @JoinColumn(name = "id", insertable = false, updatable = false)
+    private Level levelID;
```
AntonMaksimenkau: No need for an ID here

src/main/java/com/team4/testingsystem/entities/Question.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @ManyToOne
+    @JoinColumn(name = "id", insertable = false, updatable = false)
+    private Module moduleID;
```
AntonMaksimenkau: No need for an ID here

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/entities/Question.java
Darina007: point is the weight of the question from 1 to 10

src/main/java/com/team4/testingsystem/entities/Question.java
Darina007: it happened due to the fact that the isAvailable field in the database has the int type, that's why I wrote int here, do I need to fix this in the database?

P.s. of course the isAvailable method would be more logical

src/main/java/com/team4/testingsystem/entities/Question.java
AntonMaksimenkau: Didn't we decide that we are going to skip adaptive questions for now?

AntonMaksimenkau (changes requested):
Please, don't forget to create tests for your code

src/main/java/com/team4/testingsystem/entities/Question.java
Darina007: now i'm not sure about it

src/main/java/com/team4/testingsystem/entities/Question.java
AntonMaksimenkau: Yep, I think it will be much more intuitive

Darina007: ok then I'll fix it in the database and make a PR to the repository with docs

src/main/java/com/team4/testingsystem/controllers/QuestionController.java
```diff
@@ -0,0 +1,28 @@
...
+    public QuestionController(QuestionService questionService) {
+        this.questionService = questionService;
+    }
+
+    @GetMapping("/{id}")
+    public Question getQuestion(@PathVariable("id") Long id) {
+        return questionService.getQuestionById(id);
+    }
+
+    @PostMapping("/update")
```
AntonMaksimenkau: /question/update URL that updates only it's status seams confusing.
Please, use another mapping.
Or it can be done in a method that makes update operation for entire question

src/main/java/com/team4/testingsystem/entities/Level.java
```diff
@@ -0,0 +1,29 @@
+package com.team4.testingsystem.entities;
+
+import javax.persistence.*;
```
AntonMaksimenkau: Please, don't use package imports

src/main/java/com/team4/testingsystem/services/impl/LevelServiceImpl.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Autowired
+    public LevelServiceImpl(LevelRepository levelRepository) {
+        this.levelRepository = levelRepository;
+    }
+
+    @Override
+    public Level getLevelById(Long id) {
+        Level level = levelRepository.findById(id).orElse(null);
+        if (level == null) throw new NotFoundException();
```
AntonMaksimenkau: What do you think about optional.orElseThrow?

src/main/java/com/team4/testingsystem/services/impl/QuestionServiceImpl.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Override
+    public Question getQuestionById(Long id) {
+        Question question = questionRepository.findById(id).orElse(null);
+        if (question == null) throw new NotFoundException();
+        return question;
+    }
+
+    @Override
+    public Question createQuestion(String questionBody, boolean isAvailable, Long creatorUserId, Long levelID, Long moduleID) {
```
AntonMaksimenkau: Let's create a separate class instead of passing such amount of parameters

src/main/java/com/team4/testingsystem/services/impl/QuestionServiceImpl.java
```diff
@@ -0,0 +1,59 @@
...
+        Question question = new Question();
+        question.setQuestionBody(questionBody);
+        question.setAvailable(isAvailable);
+        User user = usersService.getUserById(creatorUserId);
+        question.setCreator(user);
+        Level level = levelService.getLevelById(levelID);
+        question.setLevel(level);
+        Module module = moduleService.getModuleById(moduleID);
+        question.setModule(module);
+        questionRepository.save(question);
```
AntonMaksimenkau: I really don't like an idea of such a long initialization. Especially as it is divided by getting other objects. Can we create a constructor and make it in one line?

src/main/java/com/team4/testingsystem/services/impl/QuestionServiceImpl.java
```diff
@@ -0,0 +1,59 @@
...
+        if (question == null) throw new NotFoundException();
+        return question;
+    }
+
+    @Override
+    public Question createQuestion(String questionBody, boolean isAvailable, Long creatorUserId, Long levelID, Long moduleID) {
+        Question question = new Question();
+        question.setQuestionBody(questionBody);
+        question.setAvailable(isAvailable);
+        User user = usersService.getUserById(creatorUserId);
```
AntonMaksimenkau: Please get the id from jwt token instead of passing id through parameters

src/main/resources/db/changelog/changes/schema.xml
```diff
@@ -70,13 +70,12 @@
                 <constraints primaryKey="true"/>
             </column>
             <column name="question_body" type="clob"/>
-            <column name="points" type="int"/>
```
AntonMaksimenkau, line 73: Shouldn't it be done with a change log?

src/main/resources/db/changelog/changes/schema.xml
```diff
@@ -70,13 +70,12 @@
...
             <column name="level_id" type="int">
             </column>
             <column name="module_id" type="int">
             </column>
             <column name="creator_user_id" type="int"/>
-            <column name="is_avaliable" type="boolean" defaultValue="true"/>
+            <column name="is_available" type="boolean" defaultValue="true"/>
```
AntonMaksimenkau, line 78: Code style differs from the rest of a file

AntonMaksimenkau (changes requested):
Comments added

src/main/resources/db/changelog/changes/schema.xml
AntonMaksimenkau: Shoudn't it be done with a change log?

src/main/java/com/team4/testingsystem/services/impl/QuestionServiceImpl.java
Darina007: I was told that a constructor with 3 or more parameters it is bad practice, can I use lombok for this?

AntonMaksimenkau: Any method with a big amount of parameters is bad practice.
The reason here is that you may have a mistake in the ordering of your parameters, if they have the same type.
Lombok will generate exactly the same constructor. So it is not a good fix.

At the same time, splitting the initialization is a really bad practice as well. It can be really hard to find all the places where you are using it, for example, if you need to add a new parameter.

From that perspective, I would say that it's better to violate the first rule, than the second.
Alternatively, you can create a builder for the object. I think it's one of the best fixes to the problem.

src/main/java/com/team4/testingsystem/dto/QuestionDTO.java
```diff
@@ -0,0 +1,91 @@
...
+            questionDTO.isAvailable = isAvailable;
+            return this;
+        }
+
+        public Builder creatorId(User user){
+            questionDTO.creatorId = user.getId();
+            return  this;
+        }
+
+        public Builder levelId(Level level) {
```
AntonMaksimenkau: Seams really confusing. LevelId method in builder expected to receive an Id of a level.

Alternatively, it could be a method level, which will save only id in dto. 

Anyway, builder method should define, what are they expected to receive

src/main/java/com/team4/testingsystem/dto/QuestionDTO.java
```diff
@@ -0,0 +1,91 @@
+package com.team4.testingsystem.dto;
+
+import com.team4.testingsystem.entities.Level;
+import com.team4.testingsystem.entities.Module;
+import com.team4.testingsystem.entities.User;
+
+public class QuestionDTO {
```
AntonMaksimenkau, line 5: Here, DTO means an object we are going to use for communicating with frontend. Don't think they should use ID's for level or module. I would expect them to specify them by its names

src/main/java/com/team4/testingsystem/controllers/QuestionController.java
```diff
@@ -0,0 +1,28 @@
+package com.team4.testingsystem.controllers;
+
+import com.team4.testingsystem.entities.Question;
+import com.team4.testingsystem.services.QuestionService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.web.bind.annotation.*;
```
AntonMaksimenkau: Please, stop using package import

src/main/java/com/team4/testingsystem/controllers/QuestionController.java
```diff
@@ -0,0 +1,28 @@
...
+        this.questionService = questionService;
+    }
+
+    @GetMapping("/{id}")
+    public Question getQuestion(@PathVariable("id") Long id) {
+        return questionService.getQuestionById(id);
+    }
+
+    @PostMapping("/update")
+    public Question updateQuestion(@RequestParam Long id, boolean isAvailable) {
```
AntonMaksimenkau: I think we should discuss a use case for the operation

src/main/java/com/team4/testingsystem/entities/Question.java
```diff
@@ -0,0 +1,136 @@
...
+    public Module getModule() {
+        return module;
+    }
+
+    public void setModule(Module module) {
+        this.module = module;
+    }
+
+    @Override
+    public String toString() {
```
AntonMaksimenkau: Are we going to use toString?

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/dto/QuestionDTO.java
Darina007: do i need to replace user id with user name too?

AntonMaksimenkau: Yes

src/main/java/com/team4/testingsystem/controllers/QuestionController.java
```diff
@@ -0,0 +1,50 @@
...
+        this.questionService = questionService;
+        this.questionConverter = questionConverter;
+    }
+
+    @GetMapping("/{id}")
+    public QuestionDTO getQuestion(@PathVariable("id") Long id) {
+        return questionConverter.convertToDTO(questionService.getQuestionById(id));
+    }
+
+    @PostMapping(path = "/add")
```
AntonMaksimenkau: post method with path '/'

src/main/java/com/team4/testingsystem/controllers/QuestionController.java
```diff
@@ -0,0 +1,50 @@
...
+    public QuestionDTO addQuestion(@RequestBody QuestionDTO questionDTO) {
+        return questionConverter.convertToDTO(
+                questionService.createQuestion(questionConverter.convertToEntity(questionDTO))
+        );
+    }
+
+    @PostMapping("/delete")
```
AntonMaksimenkau: delete mapping with path '/{id}'

src/main/java/com/team4/testingsystem/controllers/QuestionController.java
```diff
@@ -0,0 +1,50 @@
...
+    public void archiveQuestion(@RequestParam Long id, @RequestParam boolean isAvailable) {
+        questionService.archiveQuestion(isAvailable, id);
+    }
+
+    @PostMapping("/update/{id}")
```
AntonMaksimenkau: put mapping with path '/{id}'

src/main/java/com/team4/testingsystem/controllers/QuestionController.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @PostMapping(path = "/add")
+    public QuestionDTO addQuestion(@RequestBody QuestionDTO questionDTO) {
+        return questionConverter.convertToDTO(
+                questionService.createQuestion(questionConverter.convertToEntity(questionDTO))
+        );
+    }
+
+    @PostMapping("/delete")
+    public void archiveQuestion(@RequestParam Long id, @RequestParam boolean isAvailable) {
```
AntonMaksimenkau: Let's delete "isAvailable" from here. User deletes question. It will always be true

src/main/java/com/team4/testingsystem/controllers/QuestionController.java
```diff
@@ -0,0 +1,50 @@
...
+        questionService.archiveQuestion(isAvailable, id);
+    }
+
+    @PostMapping("/update/{id}")
+    public QuestionDTO updateQuestion(@RequestBody QuestionDTO questionDTO, @PathVariable("id") Long id) {
+        return questionConverter.convertToDTO(
+                questionService.updateQuestion(questionConverter.convertToEntity(questionDTO, id), id)
```
AntonMaksimenkau: too much-nested code. Let's make an update and return dto in different code commands;

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/repositories/QuestionRepository.java
```diff
@@ -0,0 +1,20 @@
...
+import java.util.Optional;
+
+@Repository
+public interface QuestionRepository extends CrudRepository<Question, Long> {
+    Optional<Question> findById(Long id);
+
+    Question save(Question question);
+
+    @Modifying
+    @Query(value = "update Question q set q.isAvailable = false where q.id = ?2")
```
AntonMaksimenkau: q.id = ?1

AntonMaksimenkau (changes requested):
Please, don't forget about the tests

src/main/java/com/team4/testingsystem/exceptions/NotFoundException.java
```diff
@@ -0,0 +1,8 @@
+package com.team4.testingsystem.exceptions;
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(value = HttpStatus.NOT_FOUND, reason = "Such question not found")
```
maxim12321: It seems than this exception is used not only for "question not found" reason (also levels and modules).
Should we use separate exceptions or just change the reason?)

src/test/java/com/team4/testingsystem/services/impl/LevelServiceImplTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @InjectMocks
+    private LevelServiceImpl levelService;
+
+    @Test
+    void getLevelById() {
+        Level level = EntityCreatorUtil.createLevel();
+        Mockito.when(levelRepository.findById(level.getId())).thenReturn(Optional.of(level));
+        Level result = levelService.getLevelById(level.getId());
+        Assertions.assertEquals(result, level);
```
maxim12321: Assertions are usually used as `assert...(expected, actual)`

src/test/java/com/team4/testingsystem/utils/EntityCreatorUtil.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public static Question createQuestion() {
+        return new Question.Builder()
+                .id(1L)
+                .questionBody("some text")
+                .module(createModule())
+                .level(createLevel())
+                .creator(createUser())
+                .isAvailable(true)
+                .build();
```
maxim12321: Why do we need fluent interface for questions, since we have `QuestionConverter`?)
It doesn't really make the code shorter or more clean))
Moreover, Question has fluent interface, but User, Module and Level don't... Despite they're constructed similarly
Seems a bit weird 😳

src/main/java/com/team4/testingsystem/dto/QuestionDTO.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public String getModule() {
+        return module;
+    }
+
+    public void setModule(String module) {
+        this.module = module;
+    }
+
+    public static class Builder {
```
maxim12321, line 64: Since this builder is used only to construct DTO from Question object, maybe constructor from Question is enough?

maxim12321 (changes requested):

src/main/java/com/team4/testingsystem/exceptions/NotFoundException.java
Darina007: so, you are right, i think we have no reason to use separate exceptions for such kind of exception

src/main/java/com/team4/testingsystem/dto/QuestionDTO.java
Darina007: in my opinion, using the builder is more convenient (and may be useful in the future)

src/main/java/com/team4/testingsystem/entities/Question.java
AntonMaksimenkau: private String body;

src/main/java/com/team4/testingsystem/entities/User.java
```diff
@@ -80,4 +81,12 @@ public String getLanguage() {
     public void setLanguage(String language) {
         this.language = language;
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        User user = (User) o;
+        return Objects.equals(id, user.id) && Objects.equals(name, user.name) && Objects.equals(login, user.login) && Objects.equals(password, user.password) && Objects.equals(role, user.role) && Objects.equals(language, user.language);
```
AntonMaksimenkau: Please, don't use lines bigger than 120 symbols

AntonMaksimenkau (changes requested):

src/test/java/com/team4/testingsystem/converters/QuestionConverterTest.java
```diff
@@ -0,0 +1,84 @@
...
+    @Test
+    void convertToEntityWithId() {
+        User user = EntityCreatorUtil.createUser();
+        Question question = EntityCreatorUtil.createQuestion();
+        QuestionDTO questionDTO = EntityCreatorUtil.createQuestionDto();
+        CustomUserDetails userDetails = new CustomUserDetails(user);
+        try(MockedStatic<JwtTokenUtil> jwtTokenUtilMockedStatic = Mockito.mockStatic(JwtTokenUtil.class)){
+            jwtTokenUtilMockedStatic.when(JwtTokenUtil::extractUserDetails).thenReturn(userDetails);
+            Mockito.when(questionService.getQuestionById(question.getId())).thenReturn(question);
+            Mockito.when(usersService.getUserById(JwtTokenUtil.extractUserDetails().getId())).thenReturn(user);
```
maxim12321, line 49: `JwtTokenUtil.extractUserDetails()` is mocked and returns `userDetails`
You can replace `JwtTokenUtil.extractUserDetails().getId()` with `userDetails.getId()`, or even `user.getId()` :)

maxim12321 (dismissed):

AntonMaksimenkau (approved):

AntonMaksimenkau (changes requested):
shema.xml can't be changed in a such way. Create your new changeSet

AntonMaksimenkau (approved):

maxim12321 (approved):

PR #9, opened 2021-07-10
Crud: tests

Anlord241:


src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,61 @@
...
+import com.team4.testingsystem.entities.TestEntity;
+
+import com.team4.testingsystem.services.TestsService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.*;
+
+import java.util.Optional;
+
+@Controller
```
AntonMaksimenkau: You can use @ RestController  instead of combination of @ Controller  and @ ResponseBody on each method

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @GetMapping(path = "/all")
+    public @ResponseBody
+    Iterable<TestEntity> getAllTests() {
+        return testsService.getAllTests();
+    }
+
+    @GetMapping(path = "/get")
+    public @ResponseBody
+    Optional<TestEntity> getTestById(@RequestParam long id) {
```
AntonMaksimenkau: Let's use path variables on get requests

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,61 @@
...
+    @Autowired
+    private TestsService testsService;
+
+    @GetMapping(path = "/all")
+    public @ResponseBody
+    Iterable<TestEntity> getAllTests() {
+        return testsService.getAllTests();
+    }
+
+    @GetMapping(path = "/get")
```
AntonMaksimenkau: The default rest mapping to get a single object is ${id}

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,61 @@
...
+    public @ResponseBody
+    Optional<TestEntity> getTestById(@RequestParam long id) {
+        return testsService.getTestById(id);
+    }
+
+    @PostMapping(path = "/add")
```
AntonMaksimenkau: Same here. Let's use a post request without the prefix

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,61 @@
...
+    public @ResponseBody
+    String addNewTest(@RequestParam long userId,
```
AntonMaksimenkau: User id will be transferred in jwt token. Not the params

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,61 @@
...
+                      @RequestParam String createdAt,
+                      @RequestParam String updatedAt,
+                      @RequestParam String startedAt,
+                      @RequestParam String finishedAt,
+                      @RequestParam String status,
+                      @RequestParam int evaluation) {
```
AntonMaksimenkau: Too many params here. Let's create a separate class and user it as a param

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,61 @@
...
+
+        return testsService.addNewTest(userId, createdAt, updatedAt, startedAt, finishedAt, status, evaluation);
+    }
+
+    @PostMapping(path="/update")
```
AntonMaksimenkau: Let's use put mapping with test id in request param

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,61 @@
...
+    public @ResponseBody String updateTest (@RequestParam long id,
+                                            @RequestParam long userId,
+                                            @RequestParam String createdAt,
+                                            @RequestParam String updatedAt,
+                                            @RequestParam String startedAt,
+                                            @RequestParam String finishedAt,
+                                            @RequestParam String status,
+                                            @RequestParam int evaluation) {
```
AntonMaksimenkau: Let's use one class instead such amount of parameters

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,61 @@
...
+                      @RequestParam int evaluation) {
+
+        return testsService.addNewTest(userId, createdAt, updatedAt, startedAt, finishedAt, status, evaluation);
+    }
+
+    @PostMapping(path="/update")
+    public @ResponseBody String updateTest (@RequestParam long id,
+                                            @RequestParam long userId,
+                                            @RequestParam String createdAt,
+                                            @RequestParam String updatedAt,
```
AntonMaksimenkau: You definitely shouldn't receive create/update time from user. We can get a current time by ourselves

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,61 @@
...
+                                            @RequestParam String startedAt,
+                                            @RequestParam String finishedAt,
```
AntonMaksimenkau: Let's just add start and finish methods. BTW, it can be done in a separate task

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,61 @@
...
+                                            @RequestParam String status,
```
AntonMaksimenkau: Should user specify the status of a test?

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,61 @@
...
+                                            @RequestParam int evaluation) {
+
+       return testsService.updateTest(id, userId, createdAt, updatedAt, startedAt, finishedAt, status, evaluation);
+    }
+
+    @DeleteMapping(path="/remove")
```
AntonMaksimenkau: Same here. DeleteMapping with path=/${id}

src/main/java/com/team4/testingsystem/entities/TestEntity.java
```diff
@@ -0,0 +1,105 @@
...
+public class TestEntity {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @Column(name = "id")
+    private Long id;
+
+
+    @ManyToOne
+    @JsonSerialize(using = UserSerializer.class)
```
AntonMaksimenkau: Why do we need to serialize user?

src/main/java/com/team4/testingsystem/services/TestsService.java
```diff
@@ -0,0 +1,127 @@
...
+    public Iterable<TestEntity> getAllTests() {
+        return testsRepository.findAll();
+    }
+
+    public Optional<TestEntity> getTestById(long id) {
+        try {
+            if (!testsRepository.existsById(id)) {
+                throw new NoSuchElementException("Test not found");
+            }
+            return testsRepository.findById(id);
```
AntonMaksimenkau: Why can't we just get a test and throw an exception if it wasn't find. Why do we need a separate request with verifying the test exists?

src/main/java/com/team4/testingsystem/services/TestsService.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+        catch (NoSuchElementException e) {
+            System.err.println(e.getMessage());
```
AntonMaksimenkau: Use logger instead of writing in System.*

src/main/java/com/team4/testingsystem/services/TestsService.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+        return null;
```
AntonMaksimenkau: Please, just throw an exception. We don't need nulls in our application

src/main/java/com/team4/testingsystem/services/TestsService.java
```diff
@@ -0,0 +1,127 @@
...
+            if (!usersRepository.existsById(userId)) {
+                throw new NoSuchElementException("User not found");
+            }
+            testEntity.setUser(usersRepository.findById(userId).get());
+            testEntity.setCreatedAt(LocalDateTime.parse(createdAt));
+            testEntity.setUpdatedAt(LocalDateTime.parse(updatedAt));
+            testEntity.setStartedAt(LocalDateTime.parse(startedAt));
+            testEntity.setFinishedAt(LocalDateTime.parse(finishedAt));
+            testEntity.setStatus(status);
+            testEntity.setEvaluation(evaluation);
```
AntonMaksimenkau: 8 rows for initializing a single object. As for me it's really weird

src/main/java/com/team4/testingsystem/services/TestsService.java
```diff
@@ -0,0 +1,127 @@
...
+
+            testsRepository.save(testEntity);
+            return "Success";
```
AntonMaksimenkau: Just throw an exception if something went wrong. Return codes are bad practice

src/main/java/com/team4/testingsystem/services/TestsService.java
```diff
@@ -0,0 +1,127 @@
...
+
+        } catch (NoSuchElementException e) {
+            System.err.println(e.getMessage());
+            return "Fail";
+        }
+
+    }
+
+
+    public String updateTest (long id,
```
AntonMaksimenkau: Get test by ID, get all tests, update test... You are already inside a test service. Test is already in a context here

src/test/java/com/team4/testingsystem/services/TestsServiceTest.java
```diff
@@ -0,0 +1,158 @@
...
+
+    @InjectMocks
+    TestsService testsService;
+
+    @Test
+    void successfulGetting() {
+
+        Mockito.when(testsRepository.existsById(1L)).thenReturn(true);
+        Assertions.assertNotNull(testsService.getTestById(1));
+
```
AntonMaksimenkau, line 55: Pretty strange test. According to its name, it should verify that you receive an object from the repo. 
Please, also check the other tests with the same reason

src/test/java/com/team4/testingsystem/services/TestsServiceTest.java
```diff
@@ -0,0 +1,158 @@
...
+                DATE_FOR_TESTS,
+                DATE_FOR_TESTS,
+                DATE_FOR_TESTS, 42),"Success");
+
+    }
+
+    @Test
+    void failedAddingTest() {
+
+        //User doesn't exist
```
AntonMaksimenkau: As far as id will be taken from the signed jwt token, we can be pretty sure the user exists

src/test/java/com/team4/testingsystem/services/TestsServiceTest.java
```diff
@@ -0,0 +1,158 @@
...
+        Mockito.when(usersRepository.findById(1L)).thenReturn(Optional.ofNullable(user));
+
+        Mockito.when(testsRepository.findById(1L)).thenReturn(Optional.ofNullable(testEntity));
+
+        Assertions.assertEquals(testsService.updateTest(1, 1,
+                DATE_FOR_TESTS,
+                DATE_FOR_TESTS,
+                DATE_FOR_TESTS,
+                DATE_FOR_TESTS,
+                DATE_FOR_TESTS, 42),"Success");
```
AntonMaksimenkau: I really don't like an idea to put 5 identical dates to different fields. It decreases the truly tested lines and just covers line without actual verifying

src/test/java/com/team4/testingsystem/services/TestsServiceTest.java
```diff
@@ -0,0 +1,158 @@
...
+
+
+    @Test
+    void failedUpdatingTestTestFail() {
+
+        //Test doesn't exist
+        Mockito.when(testsRepository.existsById(42L)).thenReturn(false);
+
+
+        Assertions.assertEquals(testsService.updateTest(42,1,
```
AntonMaksimenkau: AssertEquals parameters signature is (expected, actual). Please, swap them in your tests

AntonMaksimenkau (changes requested):
Please add the tests for all the classes.

And check other comments

src/main/java/com/team4/testingsystem/entities/TestEntity.java
```diff
@@ -0,0 +1,105 @@
...
+
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.team4.testingsystem.serializers.UserSerializer;
+
+import javax.persistence.*;
+import java.time.LocalDateTime;
+
+@Entity
+@Table(name = "test")
+public class TestEntity {
```
AntonMaksimenkau: Why do we need entity postfix in entity package?

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,76 @@
...
+@RequestMapping(path="/tests")
+public class TestsController {
+
+    @Autowired
+    private TestsService testsService;
+
+    @Autowired
+    private JwtTokenUtil jwtTokenUtil;
+
+    private long getUserId() {
```
AntonMaksimenkau: https://github.com/AntonMaksimenkau/team4_backend/blob/abe8ef1793e6b97eaf0ae9f89512d34a3090cd87/src/main/java/com/team4/testingsystem/utils/jwt/JwtTokenUtil.java#L84 here is a better solution for getting data from token

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,76 @@
...
+        HttpServletRequest request = ((ServletRequestAttributes) RequestContextHolder.currentRequestAttributes()).getRequest();
+        String token = request.getHeader("Authorization").split(" ")[1];
+        return jwtTokenUtil.extractUserId(token);
+    }
+
+    @GetMapping(path = "/all")
```
AntonMaksimenkau: path = /

src/main/java/com/team4/testingsystem/dto/TestsRequest.java
```diff
@@ -0,0 +1,84 @@
...
+        return new TestsRequest().new Builder();
+    }
+
+    public class Builder {
+
+        private Builder() {
+        }
+
+        public Builder setCreatedAt(String createdAt){
+            TestsRequest.this.createdAt = createdAt;
```
AntonMaksimenkau: Pretty strange constructions. Can it be simplified?

src/main/java/com/team4/testingsystem/entities/Test.java
```diff
@@ -0,0 +1,144 @@
...
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @Column(name = "id")
+    private Long id;
+
+
+    @ManyToOne
+    @JsonIgnore
+    @JoinColumn(name = "user_id", nullable = false, referencedColumnName = "id", updatable = false)
```
AntonMaksimenkau: Why updatable false?

src/main/java/com/team4/testingsystem/services/TestsService.java
```diff
@@ -0,0 +1,90 @@
...
+        this.usersRepository = usersRepository;
+    }
+
+
+    public Iterable<Test> getAll() {
+        return testsRepository.findAll();
+    }
+
+    public Test getById(long id) {
+            if (!testsRepository.existsById(id)) {
```
AntonMaksimenkau: Why can't we just get the object and throw exception if something wrong? Why do we need the extra validation?

src/main/java/com/team4/testingsystem/services/TestsService.java
```diff
@@ -0,0 +1,90 @@
...
+                throw new NoSuchElementException("Test not found");
+            }
+        return testsRepository.findById(id).get();
+    }
+
+    public long create(long userId) {
```
AntonMaksimenkau: Create method that accepts long seams confusing. Can we somehow rename it to show the logic?

src/main/java/com/team4/testingsystem/services/TestsService.java
```diff
@@ -0,0 +1,90 @@
...
+                .setCreatedAt(LocalDateTime.now())
+                .setStatus("NOT_STARTED")
+                .build();
+        testsRepository.save(test);
+        long id = test.getId();
+        return id;
+    }
+
+    public void start(long id){
+        Test test = getById(id);
```
AntonMaksimenkau: Why do we need to get the test to update it's fields?

src/test/java/com/team4/testingsystem/services/TestsServiceTest.java
AntonMaksimenkau: The test should call a method, receive the response and validate, if it is the sane data we expected

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/dto/TestsRequest.java
Anlord241: I forgot to delete this file :) It isn't used anywhere

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Autowired
+    private TestsService testsService;
+
+    @Autowired
+    private JwtTokenUtil jwtTokenUtil;
+
+    private long getUserId() {
+
+ return ((CustomUserDetails) SecurityContextHolder.getContext().getAuthentication().getPrincipal()).getId();
```
AntonMaksimenkau: Let's store it in the same file as the PR I've sent. Otherwise we will have lots of copies of the same code

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,73 @@
...
+    public void startAssigned(@PathVariable("testId") long testId) {
+        testsService.start(testId);
+    }
+
+    @PostMapping(path = "/finish/{testId}")
+    public void finish(@PathVariable("testId") long testId, @RequestParam int evaluation) {
+        testsService.finish(testId, evaluation);
+    }
+
+    @PutMapping(path="/update/{testId}")
```
AntonMaksimenkau: path="/{testId}"

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,73 @@
...
+    public void updateEvaluation(@PathVariable("testId") long testId, @RequestParam int evaluation) {
```
AntonMaksimenkau, line 59: do we need an update endpoint only for evaluation

src/main/java/com/team4/testingsystem/entities/Test.java
```diff
@@ -0,0 +1,144 @@
...
+            Test.this.startedAt = startedAt;
+            return this;
+        }
+
+        public Builder setFinishedAt(LocalDateTime finishedAt) {
+            Test.this.finishedAt = finishedAt;
+            return this;
+        }
+
+        public Builder setStatus(String status) {
```
AntonMaksimenkau: I think, the status can be enum

src/main/java/com/team4/testingsystem/services/TestsService.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    public Test getById(long id) {
+        return testsRepository.findById(id).get();
+    }
+
+    public long createForUser(long userId) {
+
+        if (!usersRepository.existsById(userId)){
+            throw  new NoSuchElementException("User not found");
```
AntonMaksimenkau: Are you catching the errors and provide one with a meaningful status code?

src/main/java/com/team4/testingsystem/services/TestsService.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        Test test = Test.newBuilder()
+                .setUser(usersRepository.findById(userId).get())
```
AntonMaksimenkau: 2 requests for one user. Let's try to get it once

src/main/java/com/team4/testingsystem/utils/jwt/JwtTokenUtil.java
```diff
@@ -27,6 +27,10 @@
         }
     }
 
+    public long extractUserId(String token) throws JwtException {
+        return Long.parseLong(extractClaim(token, Claims::getId));
```
AntonMaksimenkau: Do we actually store user id in the claims? And why do you need extract it that way?

src/test/java/com/team4/testingsystem/services/TestsServiceTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+        Assertions.assertThrows(NoSuchElementException.class, ()->testsService.createForUser(42L));
+    }
+
+    @org.junit.jupiter.api.Test
+    void startSuccess() {
+
+        Mockito.when(testsRepository.existsById(1L)).thenReturn(true);
+
+        Assertions.assertDoesNotThrow((()->testsService.start(1L)));
```
AntonMaksimenkau: Please, verify that the right method with a proper parameters were called

AntonMaksimenkau (changes requested):

AntonMaksimenkau:
Please, cover your code with tests

src/main/java/com/team4/testingsystem/entities/Test.java
```diff
@@ -0,0 +1,149 @@
...
+
+    public Builder builder() {
+        return this.new Builder();
+    }
+
+    public static Builder newBuilder() {
+        return new Test().new Builder();
+    }
+
+    public class Builder {
```
maxim12321: We have two PR's with different builder implementation. I think we should choose one of them.
As for me, [this builder](https://github.com/AntonMaksimenkau/team4_backend/blob/crud-for-questions/src/main/java/com/team4/testingsystem/entities/Question.java#L83) looks better, more clean.

src/main/java/com/team4/testingsystem/entities/User.java
```diff
@@ -11,12 +14,16 @@
...
-public class User {
+public class  User {
+
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     @Column(name = "id")
     private Long id;
 
+    @OneToMany(mappedBy = "user")
+    private Collection<Test> testEntities;
```
maxim12321: Let's name it `tests` instead of `testEntities`

src/main/java/com/team4/testingsystem/services/TestsService.java
```diff
@@ -0,0 +1,94 @@
...
+    public Iterable<Test> getAll() {
+        return testsRepository.findAll();
+    }
+
+    public Test getById(long id) {
+        try {
+            return testsRepository.findById(id).get();
+        } catch (Exception e) {
+            throw new TestNotFoundException();
+        }
```
maxim12321: This can be simplified:
```
return testsRepository.findById(id).orElseThrow(TestNotFoundException::new);
```

src/main/java/com/team4/testingsystem/services/TestsService.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    public long createForUser(long userId) {
+
+       try {
+           Test test = Test.newBuilder()
+                   .setUser(usersRepository.findById(userId).get())
```
maxim12321: Again, it can be written as follows:
```
User user = usersRepository.findById(userId).orElseThrow(UserNotFoundException::new);
```
And then:
```
.setUser(user)
```

src/main/java/com/team4/testingsystem/services/TestsService.java
```diff
@@ -0,0 +1,94 @@
...
+                   .setCreatedAt(LocalDateTime.now())
+                   .setStatus(Status.NOT_STARTED)
+                   .build();
+
+           testsRepository.save(test);
+
+           long id = test.getId();
+           return id;
```
maxim12321: Why not `return test.getId()`?)

src/main/java/com/team4/testingsystem/services/TestsService.java
```diff
@@ -0,0 +1,94 @@
...
+
+        testsRepository.updateEvaluation(LocalDateTime.now(), newEvaluation, id);
+    }
+
+    public void removeById(long id) {
+        try {
+            testsRepository.deleteById(id);
+        } catch (Exception e) {
+            throw new TestNotFoundException();
+        }
```
maxim12321: All this try-catch constructions are strange... I think solution like this can be better:
```
if (!testsRepository.existsById(id)) {
}
testsRepository.deleteById(id);
```

src/test/java/com/team4/testingsystem/services/TestsServiceTest.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    @org.junit.jupiter.api.Test
+    void createSuccess() {
+
+
+        Mockito.when(usersRepository.findById(1L)).thenReturn(Optional.of(user));
+
+        MockedStatic<Test> testMockedStatic = Mockito.mockStatic(Test.class);
+        testMockedStatic.when(Test::newBuilder).thenReturn(builder);
```
maxim12321: Mockito in it's documentation recommends to define the scope within a try-with-resources construct for [mocking static methods](https://javadoc.io/static/org.mockito/mockito-core/3.11.2/org/mockito/Mockito.html#static_mocks).
So, static mocks can be used like this:
```
try (MockedStatic<Test> mockTest = Mockito.mockStatic(Test.class)) {
    <use mockTest in this scope>
}
```

src/test/java/com/team4/testingsystem/services/TestsServiceTest.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    @org.junit.jupiter.api.Test
+    void startSuccess() {
+
+        Mockito.when(testsRepository.existsById(1L)).thenReturn(true);
+
+        testsService.start(1L);
+
+        verify(testsRepository, times(1)).start(any(LocalDateTime.class), anyLong());
```
maxim12321: You may not specify `times(1)`, it's used by default.

maxim12321 (changes requested):

src/test/java/com/team4/testingsystem/services/TestsServiceTest.java
```diff
@@ -82,30 +83,26 @@ void getByIdFail() {
...
+        try (MockedConstruction<Test.Builder> builderMockedConstruction = Mockito.mockConstruction(Test.Builder.class,
+                (mock, context) -> {
+                    Mockito.when(mock.user(any())).thenReturn(mock);
+                    Mockito.when(mock.createdAt(any())).thenReturn(mock);
+                    Mockito.when(mock.status(any())).thenReturn(mock);
+                    Mockito.when(mock.build()).thenReturn(test);
+                })) {
+            Mockito.when(test.getId()).thenReturn(1L);
+            Assertions.assertEquals(1L, testsService.createForUser(1L));
+        }
```
maxim12321: Let's leave static method for creating builder in `Test` class:
```
public static Builder builder() {
    return new Builder();
}
````

Then, you can use builder with `Test.builder()` instead of `new Test.Builder()`.
Moreover, this large constructor mocking can be replaced with mocking static method `builder()`, like it was before.
P.S. Mocking static method should be done with try-block :)

File: src/main/java/com/team4/testingsystem/services/TestsService.java (hunk 45 above)
Comment (AntonMaksimenkau) [reply src/main/java/com/team4/testingsystem/services/TestsService.java]: @maxim12321 
Not a better construction.
1. You are making 1 redundant request to DB
2. Between exists check and deletion commands can happen anything. Maybe it's not so trivial for deletion. For example, on create operation, you will have smth like that
2.1. Check test is not created
2.2. Try to save a test in DB (don't forget about multithreading, several instances, etc)
2.3. Catch exception in case 2.2 will fail.

So, the question here is, why do we need the 2.1 commands, if we will still need 2.2 and 2.3?

File: src/test/java/com/team4/testingsystem/services/TestsServiceTest.java (hunk 47 above)
Comment (AntonMaksimenkau) [reply src/test/java/com/team4/testingsystem/services/TestsServiceTest.java]: Yes, it's default for a long time. Anyway, specifying times here can make everything clear

File: src/main/java/com/team4/testingsystem/services/TestsService.java (hunk 43 above)
Comment (AntonMaksimenkau) [reply src/main/java/com/team4/testingsystem/services/TestsService.java]: Not "can be" but "should be". Here we may set a null

**Review by AntonMaksimenkau — CHANGES_REQUESTED** [review]
Comments upper

**Review by maxim12321 — DISMISSED** [review]

File: src/main/java/com/team4/testingsystem/controllers/TestsController.java [hunk 49]
```diff
@@ -0,0 +1,70 @@
...
+        return testsService.getById(id);
+    }
+
+    @PostMapping(path = "/assign/{userId}")
+    public long assign(@PathVariable("userId") long userId) {
+        return testsService.createForUser(userId);
+    }
+
+    @PostMapping(path = "/start")
+    public long startNotAssigned() {
```
AntonMaksimenkau, line 40: What test will be generated for user here?

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @GetMapping(path = "/{id}")
+    public Test getById(@PathVariable("id") long id) {
+        return testsService.getById(id);
+    }
+
+    @PostMapping(path = "/assign/{userId}")
+    public long assign(@PathVariable("userId") long userId) {
+        return testsService.createForUser(userId);
+    }
```
AntonMaksimenkau, line 37: Same here

AntonMaksimenkau (changes requested):
Please, add tests for your code

src/main/java/com/team4/testingsystem/controllers/TestsController.java
Anlord241: The method creates a test for the user, who wants to learn one's level of English without HR's help. The test is created and immediately starts. At the moment it saves only basic information about tests (such as createdAt and startedAt dates (createdAt = startedAt), status = STARTED, id, userId and initial evaluation = 0). All the other things, such as generating of questions, will be added after merging other crud tasks. By the way, the method will have more params (f.e supposed level of Engilsh)

src/main/java/com/team4/testingsystem/controllers/TestsController.java
Anlord241: Same, but it doesn't start created tests (used for assigning tests)

src/main/java/com/team4/testingsystem/controllers/TestsController.java
AntonMaksimenkau: Okay, lets do it like that for now.
Please, don't forget to add trello task with it's update

src/main/java/com/team4/testingsystem/services/TestsService.java
```diff
@@ -0,0 +1,87 @@
...
+        testsRepository.save(test);
+
+        return test.getId();
+    }
+
+    public void start(long id){
+
+        if (!testsRepository.existsById(id)){
+            throw new TestNotFoundException();
+        }
```
AntonMaksimenkau, line 56: Modifying operations in jpa can return int - amount of changed rows.
Let's use our instead of duplicating the requests

AntonMaksimenkau (changes requested):

maxim12321 (dismissed):

AntonMaksimenkau (dismissed):
please, put the dependency version to properties

maxim12321 (approved):

AntonMaksimenkau (approved):

PR #12, opened 2021-07-12
Fix authentication

maxim12321:
Add DTO for authentication request, move credentials from request URL to request body.
`/name` endpoint now returns the name of the user, not the login.
Fix some other issues.

### DIFF
--- a/src/main/java/com/team4/testingsystem/controllers/AuthenticationController.java
+++ b/src/main/java/com/team4/testingsystem/controllers/AuthenticationController.java
@@ -4,9 +4,6 @@ import com.team4.testingsystem.dto.AuthenticationRequest;
 import com.team4.testingsystem.security.CustomUserDetails;
 import com.team4.testingsystem.services.AuthenticationService;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.http.HttpStatus;
-import org.springframework.http.ResponseEntity;
-import org.springframework.security.authentication.BadCredentialsException;
 import org.springframework.security.core.Authentication;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PostMapping;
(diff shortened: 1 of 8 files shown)


src/main/java/com/team4/testingsystem/security/CustomUserDetails.java
```diff
@@ -31,7 +33,7 @@ public String getPassword() {
 
     @Override
     public String getUsername() {
-        return username;
+        return login;
```
AntonMaksimenkau, line 36: Why getUserName returns login?

src/main/java/com/team4/testingsystem/security/CustomUserDetails.java
```diff
@@ -9,12 +9,14 @@
 import java.util.Collections;
 
 public class CustomUserDetails implements UserDetails {
-    private final String username;
+    private final String name;
```
AntonMaksimenkau, line 12: Do we need to add new fields at the current OR?

src/main/java/com/team4/testingsystem/controllers/AuthenticationController.java
```diff
@@ -25,16 +25,19 @@ public AuthenticationController(AuthenticationService authenticationService) {
...
-        if (jwt_token.isEmpty()) {
+    public ResponseEntity<?> login(@RequestBody AuthenticationRequest credentials) {
+        try {
+            String jwtToken = authenticationService.createAuthenticationToken(
+                    credentials.getLogin(),
+                    credentials.getPassword()
+            );
+            return ResponseEntity.ok(jwtToken);
+        } catch (BadCredentialsException e) {
             return ResponseEntity.status(HttpStatus.UNAUTHORIZED).build();
```
AntonMaksimenkau: Please, create a corresponding exception with appropriate status code.
Such constructions seams weird.

[Here is a good example](https://www.github.com/AntonMaksimenkau/team4_backend/tree/crud-for-questions/src%2Fmain%2Fjava%2Fcom%2Fteam4%2Ftestingsystem%2Fexceptions%2FNotFoundException.java)

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/security/CustomUserDetails.java
maxim12321: `getUsername()` method is overrided from `UserDetails`, so name cannot be changed.
I used `login` instead of `username` just not to mistake `username` for `name`.
Should I change `login` back to `username`?

AntonMaksimenkau: Good point. Leave it as is

AntonMaksimenkau (approved):

Anlord241 (approved):

Darina007 (approved):

PR #21, opened 2021-07-15
Add CRUD services for files

maxim12321:
Provide two implementations of file storage:
* **FileSystem** - files are stored in temp directory
* **AmazonS3** - files are uploaded to Amazon S3 service

Amazon S3 is enabled only when `release` spring profile is active.

### DIFF
--- a/src/main/java/com/team4/testingsystem/exceptions/FileSavingFailedException.java
+++ b/src/main/java/com/team4/testingsystem/exceptions/FileSavingFailedException.java
@@ -1,0 +1,4 @@
+package com.team4.testingsystem.exceptions;
+
+public class FileSavingFailedException extends RuntimeException {
+}
--- a/src/main/java/com/team4/testingsystem/exceptions/FileSavingFailedException.java
+++ b/src/main/java/com/team4/testingsystem/exceptions/FileSavingFailedException.java
@@ -1,4 +1,8 @@
 package com.team4.testingsystem.exceptions;
 
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(value = HttpStatus.INTERNAL_SERVER_ERROR, reason = "File saving failed")
 public class FileSavingFailedException extends RuntimeException {
 }
--- a/src/main/java/com/team4/testingsystem/repositories/AmazonS3Repository.java
+++ b/src/main/java/com/team4/testingsystem/repositories/AmazonS3Repository.java
@@ -1,0 +1,61 @@
+package com.team4.testingsystem.repositories;
+
+import com.amazonaws.AmazonServiceException;
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.model.S3Object;
+import com.amazonaws.services.s3.model.S3ObjectInputStream;
+import com.amazonaws.util.IOUtils;
+import com.team4.testingsystem.exceptions.FileLoadingFailedException;
+import com.team4.testingsystem.exceptions.FileSavingFailedException;
+import org.apache.commons.io.FileUtils;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Profile;
+import org.springframework.core.io.ByteArrayResource;
+import org.springframework.core.io.Resource;
+import org.springframework.stereotype.Repository;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+@Profile("release")
+@Repository
+public class AmazonS3Repository implements FilesRepository {
+    private final AmazonS3 amazonS3;
+
+    @Value("${cloud.s3.bucket-name}")
+    private String bucketName;
+
+    @Autowired
+    public AmazonS3Repository(AmazonS3 amazonS3) {
+        this.amazonS3 = amazonS3;
+    }
+
+    @Override
+    public void save(String fileName, Resource file) throws FileSavingFailedException {
+        try {
+            Path tempFilePath = Files.createTempFile(null, file.getFilename());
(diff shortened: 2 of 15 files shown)


src/main/java/com/team4/testingsystem/repositories/AmazonS3Repository.java
```diff
@@ -0,0 +1,75 @@
...
+    public void save(String fileName, Resource file) throws FileSavingFailedException {
+        try {
+            Path tempFilePath = createTempFile(fileName);
+            File tempFile = new File(tempFilePath.toString());
+
+            FileUtils.copyInputStreamToFile(file.getInputStream(), tempFile);
+            amazonS3.putObject(bucketName, fileName, tempFile);
+            tempFile.deleteOnExit();
+        } catch (IOException | AmazonServiceException e) {
+            throw new FileSavingFailedException();
```
AntonMaksimenkau, line 52: Please, don't forget to delete a file once again, if we already created it, but didn't delete it because of an exception

src/main/java/com/team4/testingsystem/repositories/AmazonS3Repository.java
```diff
@@ -0,0 +1,75 @@
...
+        this.amazonS3 = amazonS3;
+    }
+
+    @Override
+    public void save(String fileName, Resource file) throws FileSavingFailedException {
+        try {
+            Path tempFilePath = createTempFile(fileName);
+            File tempFile = new File(tempFilePath.toString());
+
+            FileUtils.copyInputStreamToFile(file.getInputStream(), tempFile);
```
AntonMaksimenkau, line 49: I don't like an idea to save entire file on local heroku storage. Actually, I'm not sure we are able to write there and if we are - if they have enough space for our files.

Anyway, let's try that option for now. If something will work wrong - we could rewrite it (for ex. with multipart uploads)

src/main/java/com/team4/testingsystem/exceptions/FileSavingFailedException.java
```diff
@@ -0,0 +1,4 @@
+package com.team4.testingsystem.exceptions;
+
+public class FileSavingFailedException extends RuntimeException {
```
AntonMaksimenkau, line 7: Don't forget to specify error codes for your exceptions

src/main/java/com/team4/testingsystem/repositories/AmazonS3Repository.java
```diff
@@ -0,0 +1,75 @@
...
+import org.springframework.core.io.Resource;
+import org.springframework.stereotype.Repository;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+@Profile("release")
+@Repository
```
AntonMaksimenkau: I think it should be a kind of service.
@ Repository make a wrapper for translating JDBC exceptions... Definitely, we don't need it here

AntonMaksimenkau (changes requested):
So, you decided to use s3.
Is it free for our load?

src/main/java/com/team4/testingsystem/repositories/AmazonS3Repository.java
maxim12321: There's another option, `putObject` can be used with `InputStream` directly, but then `ObjectMetadata` should be passed (like content-length and so on).
It seems better, I can search how to implement this option.

src/main/java/com/team4/testingsystem/repositories/AmazonS3Repository.java
maxim12321: Maybe just `@Component` then?

AntonMaksimenkau: Why a component?

src/main/java/com/team4/testingsystem/repositories/AmazonS3Repository.java
AntonMaksimenkau: I've ask the frontend guys if they always can receive such info.
Anyway, we can try to use the solution for now and update it in case something will go wrong

src/main/java/com/team4/testingsystem/repositories/AmazonS3Repository.java
maxim12321: `FileStorage` is a service, and it uses this repositories. 
Hence, I think that they shouldn't be annotated with `@Service`, so only `@Component` can be used.

src/main/java/com/team4/testingsystem/repositories/AmazonS3Repository.java
AntonMaksimenkau: They are not sure if the can it. Do, we can't use it

src/main/java/com/team4/testingsystem/repositories/AmazonS3Repository.java
AntonMaksimenkau: Service can use another service

src/main/java/com/team4/testingsystem/repositories/AmazonS3Repository.java
maxim12321: Why we can't?
Documentation says that we must specify only `content-length`, that we can get from `MultipartFile`.
So, maybe we can pass content length as a parameter?

AntonMaksimenkau: put object requires content-length that frontend may not be able to provide.
Buffering all the files seems weird and could cause OOM exceptions.

The multipart upload will force us to make our own buffer and send the chunks. If the file will be smaller then the buffer - we can just post it as well

AntonMaksimenkau (dismissed):

AntonMaksimenkau (approved):

Anlord241 (approved):

PR #26, opened 2021-07-20
Fix deploy and configure cross-origin requests

maxim12321:


src/main/resources/application-release.properties
```diff
@@ -1,7 +1,10 @@
 # Database credentials
-spring.datasource.url=${CLEARDB_DATABASE_URL}
-spring.datasource.username=${CLEARDB_DATABASE_USERNAME}
-spring.datasource.password=${CLEARDB_DATABASE_PASSWORD}
+spring.datasource.url=jdbc:${AWS_DATABASE_URL}
+spring.datasource.username=${AWS_DATABASE_USERNAME}
+spring.datasource.password=${AWS_DATABASE_PASSWORD}
+
+spring.datasource.driver-class-name=org.postgresql.Driver
+spring.jpa.database-platform=org.hibernate.dialect.PostgreSQLDialect
```
AntonMaksimenkau: should we really specify driver and dialect? I think it can be found automatically from the url

AntonMaksimenkau (dismissed):

Darina007 (dismissed):

src/main/resources/db/changelog/changes/schema.xml
```diff
@@ -184,6 +184,7 @@
     </changeSet>
 
     <changeSet id="update_status_type" author="Anlord241">
+        <validCheckSum>8:429e8ab296a5ea20f0e1bb7bf7a5f389</validCheckSum>
```
AntonMaksimenkau, line 187: is it fix for updated DB? Will it work with a new one?

AntonMaksimenkau (approved):

src/main/resources/db/changelog/changes/schema.xml
maxim12321: This tag adds one more valid checksum, so both old and new chageSets will be valid.

maxim12321: So, for a new one, it will just execute changeSet. This fix is only for updated DB, yep

Darina007 (approved):

PR #29, opened 2021-07-21
Switched jococo to pitest

Darina007:


maxim12321 (dismissed):

AntonMaksimenkau (changes requested):
there should be a configuration, that will fail build in case mutation coverage < 60%

maxim12321 (approved):

AntonMaksimenkau (changes requested):
Can't see how it should fail a build.
You need to add the plugin execution on a mvn verify stage (better solution) ot to add a step with it's execution to a maven.yml github action

Let's also add coverageThreshold to 70

AntonMaksimenkau (approved):

maxim12321 (approved):

PR #30, opened 2021-07-21
Service implementations

Anlord241:
Created implementations for some services (every service has an interface + its implementation now)


src/main/java/com/team4/testingsystem/config/WebSecurityConfiguration.java
```diff
@@ -1,7 +1,7 @@
 package com.team4.testingsystem.config;
 
 import com.team4.testingsystem.security.JwtRequestFilter;
-import com.team4.testingsystem.services.CustomUserDetailsService;
+import com.team4.testingsystem.services.impl.CustomUserDetailsService;
```
AntonMaksimenkau: All the classes in services.impl package have Impl postfix. Let's use it here as well

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -19,7 +19,7 @@
 public class TestsController {
 
     @Autowired
-    private TestsService testsService;
+    private TestsServiceImpl testsService;
```
AntonMaksimenkau: Let's use constructor injection here

AntonMaksimenkau (changes requested):
please, fix the comments

AntonMaksimenkau (approved):

maxim12321 (approved):

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -17,12 +17,11 @@
 @RestController
 @RequestMapping(path = "/tests")
 public class TestsController {
-
-
-    private TestsService testsService;
+    
+    private TestsServiceImpl testsService;
```
maxim12321: Why `Impl` here?

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -17,12 +17,11 @@
...
 
     @Autowired
-    public TestsController(TestsService testsService) {
+    public TestsController(TestsServiceImpl testsService) {
```
maxim12321: And here

maxim12321 (changes requested):
Why all dependencies now have `Impl` type? I think we should autowire interfaces, not implementations...

src/main/java/com/team4/testingsystem/services/impl/AuthenticationServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.security.authentication.AuthenticationManager;
+import org.springframework.security.authentication.BadCredentialsException;
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
+import org.springframework.stereotype.Service;
+
+@Service
+public class AuthenticationServiceImpl implements AuthenticationService {
+    private final AuthenticationManager authenticationManager;
+
+    private final CustomUserDetailsServiceImpl userDetailsService;
```
AntonMaksimenkau: still impl here

AntonMaksimenkau (changes requested):

AntonMaksimenkau (approved):

maxim12321 (changes requested):
I think that in test files should be only one `Impl` - the implementation being tested. All other `@Mock` objects also should be interfaces.

maxim12321 (approved):

PR #34, opened 2021-07-22
Add users to database

maxim12321:


Darina007 (approved):

src/main/resources/db/changelog/changes/schema.xml
```diff
@@ -262,4 +262,102 @@
...
 
+    <changeSet id="add_users" author="maxim12321">
+        <insert tableName="user_role">
+            <column name="id" valueSequenceNext="user_role_id_seq"/>
+            <column name="role_name" value="ROLE_USER"/>
+        </insert>
+        <insert tableName="user">
+            <column name="id" valueSequenceNext="user_id_seq"/>
+            <column name="role_id" valueSequenceCurrent="user_role_id_seq"/>
+            <column name="user_login" value="rus_user"/>
```
AntonMaksimenkau: I think we should use emails as the logins.
1. We could use them for email notification
2. Probably we will need to show that it works, so the emails should be real

AntonMaksimenkau (changes requested):

src/main/resources/db/changelog/changes/schema.xml
```diff
@@ -333,15 +333,15 @@
...
+            <column name="user_login" value="rus_admin@northsixty.com"/>
             <column name="user_name" value="Russian Admin"/>
             <column name="user_password" value="$2a$08$sp14HY.XnK2w.qRE7fqMA.ja2aSL5sFGMSPSPdk/nHK.2Lj4GgK.q"/>
             <column name="language" value="rus"/>
         </insert>
         <insert tableName="user">
             <column name="id" valueSequenceNext="user_id_seq"/>
             <column name="role_id" valueSequenceCurrent="user_role_id_seq"/>
-            <column name="user_login" value="eng_admin"/>
+            <column name="user_login" value="eng_admin@northsixty.com"/>
```
AntonMaksimenkau, line 344: Do we have an access to the emails?

maxim12321: Yep, the service even promises that this temp emails will be active until we delete them manually.

AntonMaksimenkau (approved):

PR #43, opened 2021-07-26
added level's and module's

Darina007:


Anlord241 (commented):
Wouldn't it be better to use A1, A2... as level names?
I think it'll be much easier to work with them
Let's leave beatiful names to frontend

AntonMaksimenkau (changes requested):
Please, add the appropriate enums for modules and levels.
Let's use a1, a2,... levels

AntonMaksimenkau (approved):

src/main/resources/db/changelog/changes/schema.xml
```diff
@@ -376,6 +376,7 @@
     </changeSet>
 
     <changeSet id="add_level" author="Darina">
+        <validCheckSum>8:0b8ffa56b61fb87ce6cf127b376a30f2</validCheckSum>
```
AntonMaksimenkau: Not a valid changeset passing.
You need to revert a created changeset and create a new one

AntonMaksimenkau (changes requested):

src/main/resources/db/changelog/changes/schema.xml
Darina007: omg, ok😰

AntonMaksimenkau (approved):

maxim12321 (approved):

PR #47, opened 2021-07-28
CRUD service for error reports

Anlord241:


src/main/java/com/team4/testingsystem/services/impl/ErrorReportsServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+
+public class ErrorReportsServiceImpl implements ErrorReportsService {
+
+
+    private ErrorReportsRepository errorReportsRepository;
+
+    private QuestionRepository questionRepository;
+
+    private TestsRepository testsRepository;
```
maxim12321: I think here services should be used. They already throw exceptions and so on.

maxim12321 (changes requested):

maxim12321 (approved):

AntonMaksimenkau (approved):

PR #52, opened 2021-07-29
Add controller for coach's grades

maxim12321:
Configure in-memory DB for integration tests, add integration tests for CoachGrade controller.

### DIFF
--- a/src/main/java/com/team4/testingsystem/controllers/CoachGradeController.java
+++ b/src/main/java/com/team4/testingsystem/controllers/CoachGradeController.java
@@ -1,0 +1,17 @@
+package com.team4.testingsystem.controllers;
+
+import com.team4.testingsystem.services.CoachGradeService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class CoachGradeController {
+    private final CoachGradeService gradeService;
+
+    @Autowired
+    public CoachGradeController(CoachGradeService gradeService) {
+        this.gradeService = gradeService;
+    }
+
+
+}
--- a/src/main/java/com/team4/testingsystem/controllers/CoachGradeController.java
+++ b/src/main/java/com/team4/testingsystem/controllers/CoachGradeController.java
@@ -27,13 +27,6 @@ public class CoachGradeController {
         this.gradeService = gradeService;
     }
 
-    @ApiOperation(value = "Get coach's grade for a single question of the test")
-    @ApiResponse(code = 404, message = "Test, question or grade not found")
-    @GetMapping("/{testId}/{questionId}")
-    public CoachGradeDTO getGrade(@PathVariable Long testId, @PathVariable Long questionId) {
-        return new CoachGradeDTO(gradeService.getGrade(testId, questionId));
-    }
-
     @ApiOperation(value = "Get all coach's grades for the test")
     @ApiResponse(code = 404, message = "Test not found")
     @GetMapping("/{testId}")
@@ -49,14 +42,14 @@ public class CoachGradeController {
             @ApiResponse(code = 409, message = "The question is already graded")
     })
     @PostMapping("/")
-    public void createGrade(@RequestBody CoachGradeDTO coachGradeDto) {
-        gradeService.createGrade(coachGradeDto);
+    public void createGrade(@RequestBody CoachGradeDTO gradeDTO) {
+        gradeService.createGrade(gradeDTO.getTestId(), gradeDTO.getQuestionId(), gradeDTO.getGrade());
     }
 
     @ApiOperation(value = "Update grading for a single question of the test")
     @ApiResponse(code = 404, message = "Test, question or grade not found")
     @PutMapping("/")
-    public void updateGrade(@RequestBody CoachGradeDTO coachGradeDto) {
-        gradeService.updateGrade(coachGradeDto);
+    public void updateGrade(@RequestBody CoachGradeDTO gradeDTO) {
+        gradeService.updateGrade(gradeDTO.getTestId(), gradeDTO.getQuestionId(), gradeDTO.getGrade());
     }
 }
--- a/src/main/java/com/team4/testingsystem/controllers/CoachGradeController.java
+++ b/src/main/java/com/team4/testingsystem/controllers/CoachGradeController.java
@@ -54,7 +54,7 @@ public class CoachGradeController {
     }
 
     @ApiOperation(value = "Update grading for a single question of the test")
-    @ApiResponse(code = 404, message = "Grade for this question doesn't exist")
+    @ApiResponse(code = 404, message = "Test, question or grade not found")
     @PutMapping("/")
     public void updateGrade(@RequestBody CoachGradeDTO coachGradeDto) {
         gradeService.updateGrade(coachGradeDto);
--- a/src/main/java/com/team4/testingsystem/controllers/CoachGradeController.java
+++ b/src/main/java/com/team4/testingsystem/controllers/CoachGradeController.java
@@ -2,6 +2,9 @@ package com.team4.testingsystem.controllers;
 
 import com.team4.testingsystem.dto.CoachGradeDTO;
 import com.team4.testingsystem.services.CoachGradeService;
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiResponse;
+import io.swagger.annotations.ApiResponses;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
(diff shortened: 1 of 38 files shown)


src/main/java/com/team4/testingsystem/controllers/CoachGradeController.java
```diff
@@ -0,0 +1,62 @@
...
+    @Autowired
+    public CoachGradeController(CoachGradeService gradeService) {
+        this.gradeService = gradeService;
+    }
+
+    @ApiOperation(value = "Get coach's grade for a single question of the test")
+    @ApiResponse(code = 404, message = "Test, question or grade not found")
+    @GetMapping("/{testId}/{questionId}")
+    public CoachGradeDTO getGrade(@PathVariable Long testId, @PathVariable Long questionId) {
+        return new CoachGradeDTO(gradeService.getGrade(testId, questionId));
```
AntonMaksimenkau: Do we really need it?

src/main/java/com/team4/testingsystem/services/CoachGradeService.java
```diff
@@ -1,15 +1,16 @@
...
+import java.util.Collection;
 
 public interface CoachGradeService {
     CoachGrade getGrade(Long testId, Long questionId);
 
-    Iterable<CoachGrade> getGradesByTest(Long testId);
+    Collection<CoachGrade> getGradesByTest(Long testId);
 
-    void createGrade(Test test, Question question, Integer grade);
+    void createGrade(CoachGradeDTO gradeDTO);
```
AntonMaksimenkau: dto is just a transfer object. Should we really pass it to a service layer?

src/test/java/com/team4/testingsystem/controllers/CoachGradeControllerIntegrationTest.java
```diff
@@ -0,0 +1,274 @@
...
+        testsRepository.save(test);
+
+        mockMvc.perform(get("/grades/{testId}/100", test.getId())
+                .with(user(userDetails)))
+                .andExpect(status().isNotFound());
+
+        Question question = EntityCreatorUtil.createQuestion(user);
+        questionRepository.save(question);
+
+        mockMvc.perform(get("/grades/{testId}/{questionId}", test.getId(), question.getId())
```
AntonMaksimenkau: please, spread the checks to the different test cases

src/test/java/com/team4/testingsystem/controllers/CoachGradeControllerIntegrationTest.java
```diff
@@ -0,0 +1,274 @@
...
+        Assertions.assertEquals(new CoachGradeDTO(grade), gradeDTO);
+
+        gradeRepository.delete(grade);
+        questionRepository.delete(question);
+        testsRepository.delete(test);
+    }
+
+    @Test
+    void getGrades() throws Exception {
+        mockMvc.perform(get("/grades/100")
```
AntonMaksimenkau: Are you sure it will not be created by a previous test?

src/test/java/com/team4/testingsystem/controllers/CoachGradeControllerIntegrationTest.java
```diff
@@ -0,0 +1,274 @@
...
+        questionRepository.delete(question2);
+        testsRepository.delete(test);
+    }
+
+    @Test
+    void createGrade() throws Exception {
+        CoachGradeDTO gradeDTO = new CoachGradeDTO();
+        gradeDTO.setTestId(100L);
+        gradeDTO.setQuestionId(100L);
+        gradeDTO.setGrade(10);
```
AntonMaksimenkau: Why not to use builder from CoachGradeDTO?

src/test/java/com/team4/testingsystem/controllers/CoachGradeControllerTest.java
```diff
@@ -0,0 +1,162 @@
...
+    private CoachGrade coachGrade;
+
+    @Mock
+    private com.team4.testingsystem.entities.Test test;
+
+    @Mock
+    private Question question;
+
+    private final Long testId = 1L;
+    private final Long questionId = 1L;
```
AntonMaksimenkau, line 43: Let's use different ids for questions, tests, etc. Collision here may cause some passing tests even though they have errors

AntonMaksimenkau (changes requested):

src/test/java/com/team4/testingsystem/controllers/CoachGradeControllerIntegrationTest.java
maxim12321: All tests delete entities they create

src/main/java/com/team4/testingsystem/controllers/CoachGradeController.java
maxim12321: Why not? I mean, it's easier to request a grade for a specific question, instead of parsing array of DTOs.
I was thinking about returning a single Integer here, instead of DTO. Maybe it will be more convenient?

AntonMaksimenkau: What is the usecase for it?

src/test/java/com/team4/testingsystem/controllers/CoachGradeControllerIntegrationTest.java
AntonMaksimenkau: If it passed successfully... Moreover, tests can be started in parallel

maxim12321: I can clear tests, questions and grades table before each test

src/main/java/com/team4/testingsystem/controllers/CoachGradeController.java
maxim12321: When user checks the results, grades for each module are displayed separately.

src/test/java/com/team4/testingsystem/controllers/CoachGradeControllerIntegrationTest.java
AntonMaksimenkau: Why can't we just use another id?

src/main/java/com/team4/testingsystem/controllers/CoachGradeController.java
AntonMaksimenkau: For each module in the same test.
Why can't we send all the results with one request?

src/test/java/com/team4/testingsystem/controllers/CoachGradeControllerIntegrationTest.java
maxim12321: Sequence that generates IDs doesn't restart, so I think it will generate different IDs. So, correctness will depend on the order of the tests?
In my opinion, clearing tables is more correct)

maxim12321: Btw, can these tests really be started in parallel? They use common DB...

src/main/java/com/team4/testingsystem/controllers/CoachGradeController.java
maxim12321: We can, and we even send...
Ok, I'll remove

src/test/java/com/team4/testingsystem/controllers/CoachGradeControllerIntegrationTest.java
AntonMaksimenkau: Both, parallel run and different order is possible

src/test/java/com/team4/testingsystem/controllers/CoachGradeControllerIntegrationTest.java
```diff
@@ -0,0 +1,253 @@
...
+
+        mockMvc.perform(post("/grades/")
+                .contentType(MediaType.APPLICATION_JSON)
+                .content(objectMapper.writeValueAsString(gradeDTO))
+                .with(user(userDetails)))
+                .andExpect(status().isOk());
+
+        Optional<CoachGrade> grade = gradeRepository.findByTestAndQuestion(test, question);
+        Assertions.assertTrue(grade.isPresent());
+        Assertions.assertEquals(gradeDTO.getGrade(), grade.get().getGrade());
```
AntonMaksimenkau, line 210: What do you expect from the test? Are you trying to get a grade through mockMvc and from the controller, and comparing the results?

Let's try to compare mockMvc result with a grade we saved before

AntonMaksimenkau (changes requested):

src/test/java/com/team4/testingsystem/controllers/CoachGradeControllerIntegrationTest.java
maxim12321: `gradeDTO` is created at the beginning of the test, with hard-coded values, and is send in `create` request. `grade` is the value from the DB, that was saved during the request.
It's `createGrade` test, so I'm checking that the grade, saved by the controller to the DB, is equal to the value `10` from the `gradeDTO`, that was in the request.

AntonMaksimenkau (approved):

Anlord241 (approved):

PR #53, opened 2021-07-29
Methods for assigning tests to coaches

Anlord241:
I think there is no need to create a separate class for these two methods, so I've added them to TestsService


src/test/java/com/team4/testingsystem/services/impl/TestsServiceImplTest.java
```diff
@@ -187,4 +187,60 @@ void removeFail() {
...
     }
 
+    @org.junit.jupiter.api.Test
+    void assignCoachSuccess() {
+
+        Mockito.when(usersService.getUserById(1L)).thenReturn(user);
+
+        Mockito.when(testsRepository.assignCoach(user,1L)).thenReturn(1);
+
+        testsService.assignCoach(1L, 1L);
```
AntonMaksimenkau: please, use different values for different proposes

src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
```diff
@@ -42,7 +41,7 @@ public Test getById(long id) {
     public long createForUser(long userId) {
 
 
-        User user = usersRepository.findById(userId).orElseThrow(UserNotFoundException::new);
+        User user = usersService.getUserById(userId);
```
AntonMaksimenkau, line 57: what will happen if a user with the id is not found?

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
Anlord241: `.getUserById(userId)` throws `UserNotFoundException`

src/test/java/com/team4/testingsystem/services/impl/TestsServiceImplTest.java
Anlord241: I use 1 as good id and 42 as bad in all my tests. I can add something like `final long GOOD_ID = 1L` and `final long BAD_ID = 42L` to avoid magic numbers

src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
AntonMaksimenkau: yeah, it is a service, sure

src/test/java/com/team4/testingsystem/services/impl/TestsServiceImplTest.java
```diff
@@ -27,6 +27,11 @@
 @ExtendWith(MockitoExtension.class)
 class TestsServiceImplTest {
 
+    final long GOOD_TEST_ID = 1L;
+    final long GOOD_USER_ID = 1L;
+    final long BAD_USER_ID = 42L;
+    final long BAD_TEST_ID = 42L;
```
AntonMaksimenkau: final long GOOD_TEST_ID = 1L;
    
    final long GOOD_USER_ID = 2L;

    final long BAD_USER_ID = 41L;

    final long BAD_TEST_ID = 42L;`

AntonMaksimenkau (changes requested):

AntonMaksimenkau (approved):

maxim12321 (approved):

PR #54, opened 2021-07-29
test generating service

Darina007:


src/main/java/com/team4/testingsystem/repositories/ContentFilesRepository.java
```diff
@@ -19,4 +22,12 @@
     @Modifying
     @Query(value = "DELETE from ContentFile cf WHERE cf.id = ?1")
     int removeById(Long id);
+
+    @Query("SELECT cf from ContentFile cf " +
+           "JOIN cf.questions q " +
+           "JOIN q.level l " +
+           "WHERE l.name = ?1 " +
+           "order by function('random') ")
+    List<ContentFile> getRandomFiles(String level, Pageable pageable);
```
AntonMaksimenkau: why do we need a list of random files?

src/main/java/com/team4/testingsystem/services/impl/TestGeneratingServiceImpl.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.Collection;
+
+@Service
+public class TestGeneratingServiceImpl implements TestGeneratingService {
+
+    private final QuestionRepository questionService;
+    private final ContentFilesServiceImpl contentFilesService;
+
+    private static final Pageable COUNT_QUESTIONS_GRAMMAR = Pageable.ofSize(12);
```
AntonMaksimenkau: lets move constants to property file

src/main/java/com/team4/testingsystem/repositories/ContentFilesRepository.java
AntonMaksimenkau: BTW, does the pagination works on CrudRepository extention with queries without its direct usage

AntonMaksimenkau (changes requested):
Any tests here?

AntonMaksimenkau (approved):

src/main/java/com/team4/testingsystem/services/impl/TestGeneratingServiceImpl.java
```diff
@@ -0,0 +1,57 @@
...
```
AntonMaksimenkau: I think, 10 questions are enough

src/main/java/com/team4/testingsystem/services/impl/TestGeneratingServiceImpl.java
```diff
@@ -0,0 +1,60 @@
...
+public class TestGeneratingServiceImpl implements TestGeneratingService {
+
+    private final QuestionRepository questionService;
+    private final ContentFilesServiceImpl contentFilesService;
+
+    @Value("${test-generating.const.question.count}")
+    private Integer count;
+
+    @Value("${test-generating.const.one-element}")
+    private Integer oneElement;
```
AntonMaksimenkau: One element seems to be technical value, rather the user defines constant. No need to read it from properties.

src/main/java/com/team4/testingsystem/services/impl/TestGeneratingServiceImpl.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Autowired
+    public TestGeneratingServiceImpl(QuestionRepository questionService,
+                                     ContentFilesServiceImpl contentFilesService) {
+        this.questionService = questionService;
+        this.contentFilesService = contentFilesService;
+    }
+
+    @Override
+    public Test generateTest(Test test) {
```
AntonMaksimenkau: Generate test method that takes test from the argument seams weird

AntonMaksimenkau (changes requested):

AntonMaksimenkau (approved):

maxim12321 (approved):

PR #55, opened 2021-07-29
Add endpoint for getting all user's tests

maxim12321:


src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -40,6 +40,11 @@ public Test getById(@PathVariable("id") long id) {
         return testsService.getById(id);
     }
 
+    @ApiOperation(value = "Get all tests assigned to the current user")
+    @GetMapping(path = "/assigned")
```
AntonMaksimenkau: We have an operation 'assigned test', so '/assigned' mapping for tests that actually returns all the tests of the user seems a bit confusing. Let's have another mapping for it.

BTW, do we really need to getAll operation here? If not, I think '/' mapping could be fine here

src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
```diff
@@ -45,15 +43,21 @@ public Test getById(long id) {
         return testsRepository.findById(id).orElseThrow(TestNotFoundException::new);
     }
 
+    @Override
+    public Iterable<Test> getByUserId(long userId) {
+        User user = usersService.getUserById(userId);
```
AntonMaksimenkau, line 48: I think, just creating a user instance with the same id will be enough

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/controllers/TestsController.java
maxim12321: Yeah, I was thinking about replacing `GET /` mapping, moreover it's marked as experimental

src/test/java/com/team4/testingsystem/services/impl/TestsServiceImplTest.java
```diff
@@ -81,12 +81,26 @@ void getByIdFail() {
 
     }
 
+    @org.junit.jupiter.api.Test
+    void getByUserIdUserNotFound() {
```
AntonMaksimenkau, line 85: It could never happen. You are taking id from the JwtTokenUtil, which is getting user details from spring security, after it verified the correctness of the data

AntonMaksimenkau (approved):

Anlord241 (approved):

mpasenko (approved):

PR #56, opened 2021-07-30
Answer service

SemyonKorostik:


src/main/java/com/team4/testingsystem/controllers/AnswerController.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/answer")
+public class AnswerController {
```
AntonMaksimenkau: Do we really need a separate controller for answers?

src/main/java/com/team4/testingsystem/services/impl/AnswerServiceImpl.java
```diff
@@ -0,0 +1,60 @@
...
+import com.team4.testingsystem.repositories.QuestionRepository;
+import com.team4.testingsystem.services.AnswerService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class AnswerServiceImpl implements AnswerService {
+
+    AnswerRepository answerRepository;
+    QuestionRepository questionRepository;
```
AntonMaksimenkau: private?

src/main/java/com/team4/testingsystem/services/impl/AnswerServiceImpl.java
```diff
@@ -0,0 +1,60 @@
...
+                .build();
+        answerRepository.save(answer);
+    }
+
+    @Override
+    public void removeById(long id) {
+        if (!answerRepository.existsById(id)) {
+            throw new AnswerNotFoundException();
+        }
+        answerRepository.deleteById(id);
```
AntonMaksimenkau: Nothing should be deleted from DB

src/main/java/com/team4/testingsystem/services/impl/AnswerServiceImpl.java
```diff
@@ -0,0 +1,60 @@
...
+        this.questionRepository = questionRepository;
+    }
+
+    @Override
+    public Answer getById(long id) {
+        return answerRepository.findById(id).orElseThrow(AnswerNotFoundException::new);
+    }
+
+    @Override
+    public void create(AnswerRequest answerRequest) {
```
AntonMaksimenkau: AnswerRequest is a DTO. They should not exist on a service layer

src/test/java/com/team4/testingsystem/services/impl/AnswerServiceImplTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @Test
+    void getByIdSuccess() {
+        Mockito.when(answerRepository.findById(answer.getId())).thenReturn(Optional.of(answer));
+        Assertions.assertEquals(answer, answerService.getById(answer.getId()));
+    }
+
+    @Test
+    void getByIdFail() {
+        Mockito.when(answerRepository.findById(10L)).thenThrow(new AnswerNotFoundException());
```
AntonMaksimenkau: repository wouldn't throw custom exceptions

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/services/impl/AnswerServiceImpl.java
maxim12321: Let's use `QuestionService` instead of repository. Then you won't need to take care of checking if question exists and throwing exception.

src/main/java/com/team4/testingsystem/services/AnswerService.java
```diff
@@ -0,0 +1,14 @@
+package com.team4.testingsystem.services;
+
+import com.team4.testingsystem.dto.AnswerRequest;
+import com.team4.testingsystem.entities.Answer;
+
+public interface AnswerService {
+    Answer getById(long id);
+
```
maxim12321: Maybe also add a method like `getAllByQuestion`? I think it's the most useful case.

maxim12321 (changes requested):

src/main/java/com/team4/testingsystem/services/AnswerService.java
AntonMaksimenkau: Actually, what we need here is an entity, that will be used to map the data from/to the DB. All the other stuff won't be used

maxim12321: Then shouldn't something like `@OneToMany` mapping be added to `Question`, so that we can get answers for a question without accessing the repository directly?

AntonMaksimenkau: Yes

src/main/java/com/team4/testingsystem/controllers/AnswerController.java
SemyonKorostik: I initially misunderstood what this service is for

AntonMaksimenkau: Actually, you just need to have an entity here. It will store answers in the DB. At the same time, all the logic connected to it will be done in another services (question and answerOptions)

AntonMaksimenkau (approved):

src/main/java/com/team4/testingsystem/services/impl/AnswerServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+import com.team4.testingsystem.entities.Question;
+import com.team4.testingsystem.exceptions.AnswerNotFoundException;
+import com.team4.testingsystem.repositories.AnswerRepository;
+import com.team4.testingsystem.services.AnswerService;
+import com.team4.testingsystem.services.QuestionService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class AnswerServiceImpl implements AnswerService {
```
AntonMaksimenkau: Not sure we really need the service

src/main/java/com/team4/testingsystem/services/AnswerService.java
```diff
@@ -0,0 +1,10 @@
+package com.team4.testingsystem.services;
+
+import com.team4.testingsystem.entities.Answer;
+import com.team4.testingsystem.entities.Question;
+
+public interface AnswerService {
```
AntonMaksimenkau: Unused interface

maxim12321 (approved):

AntonMaksimenkau (approved):

PR #49, opened 2021-07-28
Crud for options

DzmitryNovik:

PR #58, opened 2021-08-02
Error reports controller

Anlord241:
Reworked errorReportsService, added errorReportsController, added unit and integration tests for errorReportsController

### DIFF
--- a/src/main/java/com/team4/testingsystem/controllers/ErrorReportsController.java
+++ b/src/main/java/com/team4/testingsystem/controllers/ErrorReportsController.java
@@ -1,0 +1,62 @@
+package com.team4.testingsystem.controllers;
+
+import com.team4.testingsystem.dto.ErrorReportDTO;
+import com.team4.testingsystem.services.ErrorReportsService;
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiResponse;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.web.bind.annotation.DeleteMapping;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+@RestController
+@RequestMapping(path = "/error_reports")
+public class ErrorReportsController {
+
+    private ErrorReportsService errorReportsService;
+
+    @Autowired
+    public ErrorReportsController(ErrorReportsService errorReportsService) {
+        this.errorReportsService = errorReportsService;
+    }
+
+    @ApiOperation(value = "Use it to get all error reports for the test")
+    @GetMapping("/{testId}")
+    public List<ErrorReportDTO> getReports(@PathVariable long testId) {
+        return errorReportsService.getReportsByTest(testId).stream()
+                .map(ErrorReportDTO::new)
+                .collect(Collectors.toList());
+    }
+
+    @ApiOperation(value = "Is used to add an error report")
+    @ApiResponse(code = 409, message = "The question is already reported. You can update or delete the report")
+    @PostMapping(path = "/")
+    public void add(@RequestBody ErrorReportDTO errorReportDTO) {
+        errorReportsService.add(errorReportDTO.getReportBody(),
+                errorReportDTO.getQuestionId(),
+                errorReportDTO.getTestId());
+    }
+
+    @ApiOperation(value = "Use it to change report body")
+    @PutMapping(path = "/")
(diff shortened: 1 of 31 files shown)


src/main/java/com/team4/testingsystem/controllers/ErrorReportsController.java
```diff
@@ -0,0 +1,62 @@
...
+    @PostMapping(path = "/")
+    public void add(@RequestBody ErrorReportDTO errorReportDTO) {
+        errorReportsService.add(errorReportDTO.getReportBody(),
+                errorReportDTO.getQuestionId(),
+                errorReportDTO.getTestId());
+    }
+
+    @ApiOperation(value = "Use it to change report body")
+    @PutMapping(path = "/")
+    public void updateReportBody(@RequestBody ErrorReportDTO errorReportDTO) {
```
AntonMaksimenkau: A frontend is not going to store information about created reports.
Let's have a single endpoint and make the logic with creating/merging the data on our own

src/main/java/com/team4/testingsystem/controllers/ErrorReportsController.java
```diff
@@ -0,0 +1,62 @@
...
+        errorReportsService.updateReportBody(errorReportDTO.getTestId(),
+                errorReportDTO.getQuestionId(),
+                errorReportDTO.getReportBody());
+    }
+
+    @ApiOperation(value = "Use it to remove an error report from the database")
+    @DeleteMapping(path = "/")
+    public void removeByTestAndQuestion(@RequestParam long testId, @RequestParam long questionId) {
```
AntonMaksimenkau, line 48: Don't think it will ever been used

src/main/java/com/team4/testingsystem/entities/ErrorReport.java
```diff
@@ -30,6 +30,10 @@
     @JoinColumn(name = "test_id")
     private Test test;
```
AntonMaksimenkau: What do you think about the possibility to use a pair of testId and questionId as a PK instead of building meaningless IDs?

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/entities/ErrorReport.java
maxim12321: Maybe we can use [class like this](https://github.com/AntonMaksimenkau/team4_backend/pull/49/files#diff-7af3f91cccd82763ac02a1ca9c9c420fc1b93153e0cf5075a7c72a2211fb7298R9) as an ID not only for chosen option? We use a pair (question_id, test_id) almost everywhere

src/main/java/com/team4/testingsystem/controllers/ErrorReportsController.java
Anlord241: If I don't mistake, we discussed this with frontend and decided to add an ability to cancel reporting a question (f.e if the user realize the reported question is correct)

AntonMaksimenkau: Yep, you are right

src/main/java/com/team4/testingsystem/services/impl/ErrorReportsServiceImpl.java
```diff
@@ -29,34 +34,36 @@ public ErrorReportsServiceImpl(ErrorReportsRepository errorReportsRepository,
...
         errorReportsRepository.save(new ErrorReport(requestBody, question, test));
     }
 
-    @Override
-    public void updateRequestBody(long id, String newRequestBody) {
-        if (errorReportsRepository.changeReportBody(newRequestBody, id) == 0) {
+    public void updateReportBody(long testId, long questionId, String newReportBody) {
+        if (errorReportsRepository.changeReportBody(newReportBody,
+                testsService.getById(testId),
+                questionService.getQuestionById(questionId)) == 0) {
```
maxim12321: Let's move `changeReportBody` to a separate line, 'cause this `if` looks weird
Same with `remove` method)

src/test/java/com/team4/testingsystem/controllers/ErrorReportsControllerTest.java
```diff
@@ -0,0 +1,213 @@
...
+                () -> errorReportsController.getReports(BAD_TEST_ID));
+    }
+
+    @Test
+    void addSuccess() {
+        errorReportsController
+                .add(EntityCreatorUtil
+                        .createErrorReportDTO(GOOD_REPORT_BODY,
+                                GOOD_QUESTION_ID,
+                                GOOD_TEST_ID));
```
maxim12321: image

src/main/java/com/team4/testingsystem/entities/Question.java
```diff
@@ -16,7 +17,7 @@
 
 @Entity
 @Table(name = "question")
-public class Question {
+public class Question implements Serializable {
```
AntonMaksimenkau, line 20: do we really need it?

AntonMaksimenkau (approved):

maxim12321 (approved):

src/main/java/com/team4/testingsystem/entities/Question.java
Anlord241: Yes, we get the problem below without it
image

PR #60, opened 2021-08-02
Configure PITest maven plugin

maxim12321:
* Increase code coverage threshold up to 90%
* Increase mutation coverage threshold up to 75%
* Exclude classes from `config` package
* Change execution phase to `verify`, so that other plugins run before PITest


Darina007 (approved):

Anlord241 (approved):
I approve all 4 tasks

PR #66, opened 2021-08-03
Update endpoint for adding question 

Darina007:


src/main/java/com/team4/testingsystem/controllers/QuestionController.java
```diff
@@ -36,7 +41,9 @@ public QuestionDTO getQuestion(@PathVariable("id") Long id) {
     @ApiOperation(value = "Add a new question")
     @PostMapping("/")
     public QuestionDTO addQuestion(@RequestBody QuestionDTO questionDTO) {
-        Question question = questionService.createQuestion(questionConverter.convertToEntity(questionDTO));
+        Question question = questionService
+                .createQuestion(questionConverter.convertToEntity(questionDTO), questionDTO.getAnswers());
+        answerRepository.updateCorrect(questionDTO.getCorrect());
```
AntonMaksimenkau: Can we find a way to mark a response as a correct without duplicating the data.

BTW: some questions doesn't have correct answers

AntonMaksimenkau: We definitely shouldn't call the repository from the controller. There's a service layer between them

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/controllers/QuestionController.java
Darina007: Do you mean that some questions have no answers at all?

src/main/java/com/team4/testingsystem/dto/AnswerDTO.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.team4.testingsystem.entities.Answer;
+
+import java.util.Objects;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class AnswerDTO {
+    private String answer;
+    private Boolean correct;
```
maxim12321, line 11: Are we gonna use this DTO inside TestDTO and send correct answers to the user?)

Darina007: the testDTO uses a questionDTO in which the value of the correct field is not passed on creation
it will looks like 
image

src/main/java/com/team4/testingsystem/controllers/QuestionController.java
```diff
@@ -36,7 +37,11 @@ public QuestionDTO getQuestion(@PathVariable("id") Long id) {
     @ApiOperation(value = "Add a new question")
     @PostMapping("/")
     public QuestionDTO addQuestion(@RequestBody QuestionDTO questionDTO) {
-        Question question = questionService.createQuestion(questionConverter.convertToEntity(questionDTO));
+        Question question = questionService
+                .createQuestion(questionConverter.convertToEntity(questionDTO));
+        if (questionDTO.getAnswers() != null) {
+            questionService.addAnswers(question, questionDTO.getAnswers());
+        }
         return questionConverter.convertToDTO(question);
```
AntonMaksimenkau: Why do we need to convert QuestionDTO to Question and then convert the result back to DTO?

src/main/java/com/team4/testingsystem/dto/AnswerDTO.java
AntonMaksimenkau: We can get the info on creating the answers. And sometimes (but not always), we will need to send the correct flag as well

src/main/java/com/team4/testingsystem/services/impl/QuestionServiceImpl.java
```diff
@@ -28,9 +31,18 @@ public Question getById(Long id) {
 
     @Override
     public Question createQuestion(Question question) {
-        return questionRepository.save(question);
+        Question resultQuestion = questionRepository.save(question);
+        return questionRepository.save(resultQuestion);
```
AntonMaksimenkau: why do we need to save questions 2 times?

src/main/java/com/team4/testingsystem/services/impl/QuestionServiceImpl.java
```diff
@@ -28,9 +31,18 @@ public Question getById(Long id) {
...
     }
 
+    @Transactional
+    @Override
+    public Question addAnswers(Question question, List<AnswerDTO> textAnswers) {
+        List<Answer> answers = new ArrayList<>();
+        textAnswers.forEach(answerDTO -> answers.add(new Answer(answerDTO, question)));
```
AntonMaksimenkau: List<Answer> answers = textAnswers.map(answerDTO -> new Answer(answerDTO, question).collect(Collectors.toList())

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/dto/AnswerDTO.java
AntonMaksimenkau: Why do we need available flag in dto?

Darina007: maybe not needed)

AntonMaksimenkau (approved):

maxim12321 (approved):

PR #72, opened 2021-08-04
Added reason to response body

Anlord241:


Darina007 (approved):

AntonMaksimenkau (changes requested):
Pretty brave deceison. It can compromise some data.
Let's find another way

maxim12321 (changes requested):
Let's use one ControllerAdvice class for all exception handlers

AntonMaksimenkau (approved):

src/main/java/com/team4/testingsystem/exceptions/Response.java
```diff
@@ -0,0 +1,32 @@
+package com.team4.testingsystem.exceptions;
```
AntonMaksimenkau: Response doesn't seam like an exception. Let's move it somewhere

src/main/java/com/team4/testingsystem/exceptions/ControllerAdvice.java
```diff
@@ -0,0 +1,37 @@
...
+public class ControllerAdvice extends ResponseEntityExceptionHandler {
+
+    @ExceptionHandler(value = {ConflictException.class})
+    @ResponseStatus(value = HttpStatus.CONFLICT)
+    public ErrorResponse handleConflictException(ConflictException e) {
+        return new ErrorResponse(e.getMessage());
+    }
+
+    @ExceptionHandler(value = {FileOperationException.class})
+    @ResponseStatus(value = HttpStatus.INTERNAL_SERVER_ERROR)
```
maxim12321: Maybe we should send 4xx responses only?

Anlord241: What code should I use then?

Anlord241: I think problems with file saving/deleting are server errors but not client

maxim12321: `409 Conflict`? `417 Expectation Failed`?
`418 I'm a teapot` :)

maxim12321 (approved):

PR #73, opened 2021-08-04
Switch all TestController methods to DTO

maxim12321:


src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -33,26 +33,29 @@ public TestsController(TestsService testsService) {
...
     }
 
     @ApiOperation(value = "Get all tests assigned to the user")
     @GetMapping(path = "/history/{userId}")
-    public Iterable<Test> getUsersTests(@PathVariable("userId") long userId) {
-        return testsService.getByUserId(userId);
+    public List<TestDTO> getUsersTests(@PathVariable("userId") long userId) {
+        return testsService.getByUserId(userId).stream()
+                .map(TestDTO::new)
+                .collect(Collectors.toList());
```
AntonMaksimenkau: Seems like a code duplicate. 
Would you mind to move it to a single place

src/test/java/com/team4/testingsystem/controllers/TestsControllerIntegrationTest.java
```diff
@@ -73,28 +72,26 @@ void destroy() {
     @Test
     void getUsersTestsSuccess() throws Exception {
         com.team4.testingsystem.entities.Test test1 = EntityCreatorUtil.createTest(user);
+        test1.setCreatedAt(null);
         testsRepository.save(test1);
 
         com.team4.testingsystem.entities.Test test2 = EntityCreatorUtil.createTest(user);
+        test2.setCreatedAt(null);
```
AntonMaksimenkau, line 79: Let's check all the values are passed

AntonMaksimenkau (changes requested):

src/test/java/com/team4/testingsystem/controllers/TestsControllerIntegrationTest.java
maxim12321: There's a problem with comparing `LocalDateTime` after saving it to DB, so they are not equal...
This can be solved later, in PR where `createdAt` should be not null
I think we can switch to DTOs now, so that frontend can parse our responses, and fix this later :)

AntonMaksimenkau (approved):

Darina007 (approved):

PR #79, opened 2021-08-05
Service for test evaluation

SemyonKorostik:


src/main/java/com/team4/testingsystem/services/impl/TestEvaluationServiceImpl.java
```diff
@@ -0,0 +1,41 @@
...
+@Service
+public class TestEvaluationServiceImpl implements TestEvaluationService {
+    private final ChosenOptionService chosenOptionService;
+
+    @Autowired
+    public TestEvaluationServiceImpl(ChosenOptionService chosenOptionService) {
+        this.chosenOptionService = chosenOptionService;
+    }
+
+    public int getEvaluationByTest(Test test) {
```
AntonMaksimenkau, line 20: Too complex. It will be something like that: 
return choosenOptionService.getChoosenOptionByTest(test).map(ChoosenOption::getAnswer).filter(Answer::isCorrect) .count();

src/main/java/com/team4/testingsystem/services/TestEvaluationService.java
```diff
@@ -0,0 +1,7 @@
+package com.team4.testingsystem.services;
+
+import com.team4.testingsystem.entities.Test;
+
+public interface TestEvaluationService {
+    int getEvaluationByTest(Test test);
```
AntonMaksimenkau, line 6: Can't see where is it used. Please set the evaluation on finish method of test service

AntonMaksimenkau (changes requested):

AntonMaksimenkau (approved):

maxim12321 (approved):

PR #80, opened 2021-08-06
Added tests limit (3 tests per day)

Anlord241:
HRs don't affect the limit

### DIFF
--- a/src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
+++ b/src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
@@ -9,6 +9,7 @@ import com.team4.testingsystem.enums.Levels;
 import com.team4.testingsystem.enums.Status;
 import com.team4.testingsystem.exceptions.CoachAssignmentFailException;
 import com.team4.testingsystem.exceptions.TestNotFoundException;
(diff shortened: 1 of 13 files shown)


src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
```diff
@@ -71,6 +68,16 @@ public Test save(Test test) {
 
     @Override
     public long startForUser(long userId, Levels levelName) {
+        List<Test> selfStarted = getByUserId(userId)
+                .stream()
+                .filter((test) -> test.getAssignedAt() == null)
+                .filter((test) -> test.getStartedAt().isAfter(LocalDateTime.now().minusDays(1)))
+                .collect(Collectors.toList());
```
AntonMaksimenkau: I think it's better to add an appropriate request from the repository, instead of making the logic in the app.
User can have too many tests and downloading all of them can lead to a huge amount of problems

AntonMaksimenkau (changes requested):

AntonMaksimenkau (approved):

maxim12321 (approved):

Darina007 (approved):

PR #83, opened 2021-08-09
Add endpoint to get list of all users

maxim12321:


src/main/java/com/team4/testingsystem/controllers/UsersController.java
```diff
@@ -31,9 +36,23 @@ public UsersController(UsersService usersService) {
                 .collect(Collectors.toList());
     }
 
+    @GetMapping("/employees")
+    public List<UserDTO> getAllUsers() {
+        return usersService.getAll().stream()
+                .map(UserDTO::new)
+                .peek(user -> user.setAssignedTest(findAssignedTest(user)))
```
AntonMaksimenkau: What do you think about one operation with getting all the assigned tests within one request and add them to user stream?

AntonMaksimenkau (changes requested):

Darina007 (approved):

src/main/java/com/team4/testingsystem/controllers/UsersController.java
```diff
@@ -31,6 +37,18 @@ public UsersController(UsersService usersService) {
                 .collect(Collectors.toList());
     }
 
+    @GetMapping("/employees")
+    public List<UserDTO> getAllUsers() {
+        Status[] statuses = {Status.ASSIGNED};
```
AntonMaksimenkau: It is service logic. Not the controller one

AntonMaksimenkau (changes requested):
please, use services for buisness logic

src/main/java/com/team4/testingsystem/controllers/UsersController.java
```diff
@@ -17,23 +19,36 @@
...
-        return usersService.getUsersByRole(Role.COACH).stream()
-                .map(UserDTO::new)
-                .collect(Collectors.toList());
+        return convertToDTO(usersService.getUsersByRole(Role.COACH));
+    }
+
+    @GetMapping("/employees")
+    public List<UserDTO> getAllUsers() {
+        List<UserDTO> users = convertToDTO(usersService.getAll());
+        testsService.attachAssignedTests(users);
```
AntonMaksimenkau: please, move service logic outside from controller

src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
```diff
@@ -68,6 +72,15 @@ public Test getById(long id) {
         return testsRepository.getByStatuses(statuses);
     }
 
+    @Override
+    public void attachAssignedTests(List<UserDTO> users) {
```
AntonMaksimenkau: Let's return a map with id and test instead of modifying DTO on a service layer

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/controllers/UsersController.java
maxim12321: If I move attaching tests to service, than DTO will be modified on a service layer...

src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
maxim12321: If I return a map from `attach...` instead of modifying DTO, other service will use DTO, or controller will have business-logic...

maxim12321: Looks like an endless cycle 😳

src/main/java/com/team4/testingsystem/controllers/UsersController.java
maxim12321: BTW, why attaching assigned tests to DTO can't be done in the controller?

AntonMaksimenkau: Service should work with converting DTO`s to object and back. All the other logic should be made on another layer.

So, the service should make all the logic on itself and return only the result with users, that have (or don't have) corresponging tests

src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
AntonMaksimenkau: Can we have an object that will represent the user and his test?
If so, we can return it from service, and the controller can map it to dto on its layer only

maxim12321: I can use smth like `Map<User, Optional<Test>>` for the service layer

AntonMaksimenkau (approved):

src/main/java/com/team4/testingsystem/controllers/UsersController.java
```diff
@@ -17,17 +20,24 @@
...
     @GetMapping("/coaches")
     public List<UserDTO> getCoaches() {
-        return usersService.getUsersByRole(Role.COACH).stream()
-                .map(UserDTO::new)
+        return convertToDTO(usersService.getUsersByRole(Role.COACH));
+    }
+
+    @GetMapping("/employees")
+    public List<UserDTO> getAllUsers() {
+        return testsService.getUsersWithAssignedTests().stream()
```
AntonMaksimenkau: Let's rename the method to `getAllUsersAndAssignedTests`.
Current name looks like you're going to return only users, that have assigned tests

PR #84, opened 2021-08-09
endpoint for adding content file with list of questions

Darina007:


src/main/java/com/team4/testingsystem/controllers/QuestionController.java
```diff
@@ -59,4 +80,10 @@ public QuestionDTO updateQuestion(@RequestBody QuestionDTO questionDTO, @PathVar
                 .updateQuestion(questionConverter.convertToEntity(questionDTO, id), id);
         return questionConverter.convertToDTO(resultQuestion);
     }
+
+    private List<Question> convertToEntity(List<QuestionDTO> questionsDTO) {
+        List<Question> questions = new ArrayList<>();
+        questionsDTO.forEach(questionDTO -> questions.add(questionConverter.convertToEntity(questionDTO)));
```
AntonMaksimenkau: Why do you so much like the idea of adding entities from the stream to create an empty list?

AntonMaksimenkau (approved):

maxim12321 (approved):

PR #89, opened 2021-08-10
Update listening

Darina007:


src/main/java/com/team4/testingsystem/services/impl/QuestionServiceImpl.java
```diff
@@ -72,4 +78,11 @@ public Question updateQuestion(Question question, Long id) {
         return questionRepository.getQuestionsByTestId(id);
     }
 
+    @Transactional
+    @Override
+    public void archiveQuestionsByContentFileId(Long id) {
+        ContentFile contentFile = contentFilesRepository.findById(id)
```
AntonMaksimenkau, line 97: Let's archive content file

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/services/impl/QuestionServiceImpl.java
Darina007: don't we need to archive questions if we change the file?

AntonMaksimenkau: Theoretically, yes.
In practice, the listening test is generated by a chosen content file. If a file is archived, the questions will never be found.
Alternatively, you can make a single request, that will archive all the questions by its content file. But definitely we don't need to make 10 requests for it

src/main/java/com/team4/testingsystem/controllers/QuestionController.java
```diff
@@ -80,6 +80,20 @@ public ContentFileDTO addListening(@RequestPart MultipartFile file, @RequestPart
...
+                                          @RequestPart List<QuestionDTO> questions,
+                                          @PathVariable("contentFileId") Long id) {
+        if (file == null) {
+            ContentFile contentFile = contentFilesService
+                    .updateQuestions(id, convertToEntity(questions));
+            return new ContentFileDTO(contentFile);
+        }
+        String url = storageService.upload(file.getResource());
+        return new ContentFileDTO(contentFilesService.update(id, url, convertToEntity(questions)));
+    }
```
AntonMaksimenkau: Please, move the logic outside from the controller

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/services/ContentFilesService.java
```diff
@@ -11,7 +13,9 @@
 
     ContentFile getById(long id);
 
-    ContentFile add(String ulr, List<Question> questions);
+    ContentFile add(Resource file, List<Question> questions);
+
+    ContentFile update(MultipartFile file, Long id, List<Question> questions);
```
AntonMaksimenkau: It's pretty strange that 2 of the operations have different file type

src/main/java/com/team4/testingsystem/services/impl/ContentFilesServiceImpl.java
```diff
@@ -31,10 +40,26 @@ public ContentFile getById(long id) {
...
 
+
+    @Override
+    public ContentFile update(MultipartFile file, Long id, List<Question> questions) {
+        if (file == null) {
+            questionService.archiveQuestionsByContentFileId(id);
+            ContentFile contentFile = contentFilesRepository
+                    .findById(id).orElseThrow(FileNotFoundException::new);
+            contentFile.getQuestions().addAll(questions);
+            return contentFilesRepository.save(contentFile);
```
AntonMaksimenkau, line 56: Is it possible, that between `archiveQuestions` and save `ContentFile` with a new one someone will generate a test with it, or open the content file to edit?
If so, I think it's inappropriate and we should deal with it. I could suggest you 2 options:
1.  make the method `@Transactional`. If the changes will be made in a single transaction, we shouldn't have any problems with it
2. Even though we are not updating the file, we can create a new `ContentFile` with the same URL data and archive the previous one.

AntonMaksimenkau (changes requested):

AntonMaksimenkau (approved):

maxim12321 (approved):

PR #95, opened 2021-08-10
Added endpoint for deassigning tests

Anlord241:


Anlord241:
What is wrong with checks?

maxim12321:
https://www.githubstatus.com/
GitHub actions failed a bit :)

AntonMaksimenkau (approved):

src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
```diff
@@ -113,6 +113,16 @@ public long assignForUser(long userId, Levels levelName, LocalDateTime deadline)
         return test.getId();
     }
 
+    @Override
+    public void deassign(long id) {
```
maxim12321, line 117: Shouldn't we update the test status after deassign?

AntonMaksimenkau: No. the status should not be `assigned` at the point

maxim12321 (approved):

PR #101, opened 2021-08-11
Added 40 minutes timer

Anlord241:


Darina007 (approved):

src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -78,6 +78,12 @@ public ModuleGradesDTO getGrades(@PathVariable("testId") long testId) {
         return convertToDTO(testsService.getByStatuses(statuses));
     }
 
+    @ApiOperation(value = "Is used to get time left (in seconds)")
+    @GetMapping(path = "/time/{testId}")
+    public long getTimeLeft(@PathVariable("testId") long testId) {
```
AntonMaksimenkau: don't need to have a separate endpoint for it. Just set the finishTime to a test on /start requests

src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
```diff
@@ -136,14 +152,30 @@ public void deassign(long id) {
...
         }
         Test test = testGeneratingService.formTest(getById(id));
+
         save(test);
+
+        setTimer(id);
         return test;
     }
 
+    private void setTimer(long id) {
```
AntonMaksimenkau: it won't work after restarts. There can be 3 options here:
1. store the timers in DB and recover them on app start
2. find external timer for it
3. make a lazy finish

BTW: don't forget that after the time had passed, we still need to receive some frontend requests wtih the answers

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
AntonMaksimenkau: Anyway, we have some separate tasks to finish the test. You don't need to do it here

maxim12321 (commented):
Don't you think that all the work with timers can be moved from the `TestsService` to a separate service?

Anlord241:
> Don't you think that all the work with timers can be moved from the `TestsService` to a separate service?

Yes, I do (I even did so in the beginning)
But I think it's impossible because of cycles (we need to use `TestsService`  methods in `TimerService` and vice versa)

maxim12321:
Why do we need to use `TestsService` methods in `TimerService`?
It seeems that `createTimer`, `startTimer` and `startAllTimers` methods don't depend on `TestsService` and can be easily extracted to a separate service :)

Anlord241:
> Why do we need to use `TestsService` methods in `TimerService`?
> It seeems that `createTimer`, `startTimer` and `startAllTimers` methods don't depend on `TestsService` and can be easily extracted to a separate service :)

`startTimer` calls `finish` from `TestsService`
  and 
`startAllTimers`, `createTimer` call `startTimer`
___________________________________________________
It means, that all methods from `TimerService` call `finish` :)

maxim12321:
Maybe you can use something like callback function?)
So, you can pass function as a parameter for `Timer` (e.g. your `TestsService::finish` or it's wrapper), that will be called after timer is out.
IDK should we do that or not))

maxim12321 (approved):

src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
```diff
@@ -138,18 +149,55 @@ public void deassign(long id) {
 
     @Override
     public Test start(long id) {
+
         if (testsRepository.start(Instant.now(), id) == 0) {
             throw new TestNotFoundException();
         }
         Test test = testGeneratingService.formTest(getById(id));
+
+        test.setFinishTime(Instant.now().plusSeconds(42 * 60));
```
AntonMaksimenkau: now().plus(40l, ChronoUnit.MINUTES)?

AntonMaksimenkau: we decided to use 42 munutes for a timer. For the test it still should be 40 mins

src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
```diff
@@ -138,18 +149,55 @@ public void deassign(long id) {
...
+    public void startAllTimers() {
+        timerRepository.findAll().forEach(this::startTimer);
+    }
+
     @Override
-    public void finish(long id) {
+    public void finish(long id, Instant finishDate) {
         testEvaluationService.countScoreBeforeCoachCheck(getById(id));
-        testsRepository.finish(Instant.now(), id);
+        testsRepository.finish(finishDate, id);
```
AntonMaksimenkau: Should we stop timer in case user finished it?

AntonMaksimenkau (changes requested):

AntonMaksimenkau (approved):

PR #105, opened 2021-08-12
Include Lombok dependency in POM.xml + Use Lombok annotations.

DzmitryNovik:


src/main/java/com/team4/testingsystem/entities/ChosenOption.java
```diff
@@ -13,6 +17,11 @@
 
 @Entity
 @Table(name = "chosen_option")
+@NoArgsConstructor
+@AllArgsConstructor
+@Getter
+@Setter
+@EqualsAndHashCode
```
AntonMaksimenkau, line 24: Why don't you like to use `@Data` annotation instead of a list of other ones?

AntonMaksimenkau (changes requested):
Please, also use `@Builder` annotation instead a custom ones

AntonMaksimenkau:
BTW, what about the rest of the project? Are you going to make a separate PR for it?

src/test/java/com/team4/testingsystem/converters/TestConverterTest.java
```diff
@@ -98,6 +99,7 @@ void convertToDTO() {
                 chosenOptions.add(new ChosenOption(new TestQuestionID(test, question), answers.get(0)));
 
                 question.setAnswers(answers);
+                /*questions.add(question);*/
```
AntonMaksimenkau: why?

AntonMaksimenkau (changes requested):

src/test/java/com/team4/testingsystem/converters/TestConverterTest.java
DzmitryNovik: It was an attempt to fix error. This line is not needed.

AntonMaksimenkau: I'm pretty sure we need it.
Or do we have another place where we are setting the questions?

AntonMaksimenkau (approved):

maxim12321 (approved):

PR #107, opened 2021-08-12
Fix

Darina007:


src/main/java/com/team4/testingsystem/dto/TestDTO.java
```diff
@@ -20,7 +20,7 @@
     private String status;
     private UserDTO coach;
     private Map<String, List<QuestionDTO>> questions;
-    private ContentFileDTO contentFile;
+    private ListeningTopicRequest contentFile;
```
maxim12321: I think that's not `request` anymore :)

maxim12321 (changes requested):

maxim12321 (approved):

AntonMaksimenkau (approved):

PR #112, opened 2021-08-13
Add endpoint to get tests by level and user

SemyonKorostik:


src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -58,6 +58,12 @@ public TestsController(TestsService testsService,
         return convertToDTO(testsService.getByUserId(userId));
```
AntonMaksimenkau: Why not use the existing endpoint with an optional level parameter?

AntonMaksimenkau (changes requested):

AntonMaksimenkau (approved):

maxim12321 (approved):

PR #113, opened 2021-08-13
Add endpoint to get all unverified tests, assigned to current coach

maxim12321:


src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
```diff
@@ -69,6 +69,11 @@ public Test getById(long id) {
         return testsRepository.getByStatuses(statuses);
     }
 
+    @Override
+    public List<Test> getAllUnverifiedTestsByCoach(long coachId) {
+        return testsRepository.getAllByAssignedCoachAndStatus(coachId, Status.COMPLETED);
```
AntonMaksimenkau: in-verification?

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
maxim12321: This page is for coaches to start test verification. The page displays all **unverified** tests, that were assigned to the current user.

maxim12321: So, when the status is `in-verification`, the test shouldn't be displayed here)

AntonMaksimenkau: They should. The coach could leave a page and should have ability to continue the verification

maxim12321: Oh, yeah. I'll rename the method.

maxim12321: Oh, maybe I don't need to rename it :)

AntonMaksimenkau (approved):

Anlord241 (approved):

Darina007 (approved):

PR #123, opened 2021-08-16
Update names for uploading files 

maxim12321:


src/main/java/com/team4/testingsystem/services/impl/ResourceStorageServiceImpl.java
```diff
@@ -21,24 +19,34 @@ public ResourceStorageServiceImpl(FilesService filesService) {
     }
 
     @Override
-    public String upload(Resource file) throws FileSavingFailedException {
-        String fileUrl = UUID.randomUUID() + "-" + file.getFilename();
-        save(fileUrl, file);
-        return fileUrl;
+    public String upload(Resource file, Modules module, Long primaryId) {
```
AntonMaksimenkau, line 22: testId?

AntonMaksimenkau (approved):

src/main/java/com/team4/testingsystem/services/impl/ResourceStorageServiceImpl.java
maxim12321: That's not always testId, e.g. content file doesn't have testId, so primaryId is creator's id

Darina007 (approved):

PR #129, opened 2021-08-17
added endpoint to archive question

Darina007:


maxim12321 (approved):

Anlord241 (commented):
The task is already done here :)
https://github.com/AntonMaksimenkau/team4_backend/pull/125

Darina007:
> The task is already done here :)
> #125

then tasks in trello are not updated, oops)

PR #134, opened 2021-08-17
Setting verified status when coach presses submit

Anlord241:


src/main/java/com/team4/testingsystem/repositories/TestsRepository.java
```diff
@@ -52,8 +52,8 @@
 
     @Transactional
     @Modifying
-    @Query(value = "UPDATE Test t SET t.verifiedAt = ?1 where t.id = ?2")
-    int updateEvaluation(Instant updateDate, Long id);
+    @Query(value = "UPDATE Test t SET t.verifiedAt = ?1, t.status = 'VERIFIED' where t.id = ?2")
+    int update(Instant updateDate, Long id);
```
AntonMaksimenkau: let's rename it to something like 'verify'/`coachSubmit` or smth. like that. 'update' seems too unclear

AntonMaksimenkau (changes requested):

maxim12321 (approved):

AntonMaksimenkau (approved):

PR #136, opened 2021-08-17
added sort and pagination

Darina007:


src/main/java/com/team4/testingsystem/controllers/TestsController.java
```diff
@@ -54,19 +57,24 @@ public TestsController(TestsService testsService,
 
     @ApiOperation(value = "Get all tests assigned to the current user")
     @GetMapping(path = "/")
-    public List<TestInfo> getCurrentUserTests() {
-        return convertToTestInfoDTO(testsService.getByUserId(JwtTokenUtil.extractUserDetails().getId()));
+    public List<TestInfo> getCurrentUserTests(@RequestParam int pageNumb,
+                                              @RequestParam int pageSize) {
```
AntonMaksimenkau, line 59: What do you think about some default values?

AntonMaksimenkau (approved):

src/main/java/com/team4/testingsystem/controllers/TestsController.java
Darina007: on the pages there is a choice of the number of lines that will be displayed, I think that the front will send us the amount that they need

Anlord241 (approved):

PR #138, opened 2021-08-17
Restrictions: answer controller

Anlord241:
image

### DIFF
--- a/src/main/java/com/team4/testingsystem/controllers/AnswerController.java
+++ b/src/main/java/com/team4/testingsystem/controllers/AnswerController.java
@@ -4,6 +4,7 @@ import com.team4.testingsystem.enums.Modules;
 import com.team4.testingsystem.services.AnswerService;
 import io.swagger.annotations.ApiOperation;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.access.annotation.Secured;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.PostMapping;
@@ -32,6 +33,7 @@ public class AnswerController {
 
     @ApiOperation(value = "Upload essay text")
     @PostMapping("/essay/{testId}")
+    @Secured("ROLE_COACH")
     public String uploadEssay(@PathVariable Long testId, @RequestBody String text) {
         return answerService.uploadEssay(testId, text);
     }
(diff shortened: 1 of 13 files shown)


src/main/java/com/team4/testingsystem/exceptions/ControllerAdvice.java
```diff
@@ -58,4 +58,10 @@ public TestsLimitExceededResponse handleTestsLimitExceededException(TestsLimitEx
         }
         return handleExceptionInternal(e, body, new HttpHeaders(), HttpStatus.BAD_REQUEST, request);
     }
+
+    @ExceptionHandler(value = {DoNotHaveRightsException.class})
+    @ResponseStatus(value = HttpStatus.FORBIDDEN)
+    public ErrorResponse handleDoNotHaveRightsException(DoNotHaveRightsException e) {
```
AntonMaksimenkau: Why not to use [AccessControlException](https://docs.oracle.com/javase/7/docs/api/java/security/AccessControlException.html)?

src/main/java/com/team4/testingsystem/services/impl/FileAnswerServiceImpl.java
```diff
@@ -49,10 +49,22 @@ public String getUrl(Long testId, Long questionId) {
...
-        Question question = questionService.getQuestionByTestIdAndModule(testId, module);
-        return save(testId, question.getId(), url);
+
+        Question question = questionService.getQuestionByTestIdAndModule(testId, Modules.SPEAKING);
+
+        FileAnswer fileAnswer = FileAnswer.builder()
+            .id(new TestQuestionID(test, question))
+            .url(url)
+            .build();
+        return fileAnswerRepository.save(fileAnswer);
```
AntonMaksimenkau: Will it also work properly with update?

src/main/java/com/team4/testingsystem/services/impl/TestsServiceImpl.java
```diff
@@ -253,4 +254,19 @@ public void deassignCoach(long id) {
             throw new TestNotFoundException();
         }
     }
+
+    @Override
+    public void checkRights(Test test) {
```
AntonMaksimenkau: checkRights name is too wide while it checks only test owner. Let's rename to smth more concreate

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/services/impl/FileAnswerServiceImpl.java
Anlord241: Why not?
We rewrite file answer if it already exists

AntonMaksimenkau: We expected the behavior, but seams like something is broken there

AntonMaksimenkau (approved):

src/main/java/com/team4/testingsystem/controllers/AnswerController.java
```diff
@@ -32,6 +32,7 @@ public String downloadEssay(@PathVariable Long testId) {
 
     @ApiOperation(value = "Upload essay text")
     @PostMapping("/essay/{testId}")
+    @Secured("ROLE_COACH")
```
maxim12321: Maybe this restriction is for `downloadEssay`?

maxim12321 (changes requested):

maxim12321 (approved):

PR #173, opened 2021-08-21
Don't allow to start a test if there are not enough questions in the db

Anlord241:


src/main/java/com/team4/testingsystem/repositories/TestsRepository.java
```diff
@@ -55,7 +55,8 @@
             + "where t.user = ?1 "
             + "and t.isAvailable = true "
             + "and t.assignedAt is null "
-            + "and t.startedAt >= ?2 ")
+            + "and t.startedAt >= ?2 "
+            + "and t.isAvailable = true ")
```
AntonMaksimenkau: Is available duplicate constrain

src/main/java/com/team4/testingsystem/services/impl/TestGeneratingServiceImpl.java
```diff
@@ -43,19 +45,31 @@ public Test formTest(Test test) {
...
-        questions.forEach(test::setQuestion);
+
+        try {
+            ContentFile contentFile = contentFilesService
+                    .getRandomContentFile(test.getLevel().getName());
+
+            Collection<Question> questions = questionRepository
+                    .getRandomQuestionByContentFile(contentFile.getId(), Pageable.ofSize(count));
+
+            questions.forEach(test::setQuestion);
```
AntonMaksimenkau: Is it work like addQuestion? Set question seams a bit strange here. Anyway, can we add all the questions instead adding them one by one?

AntonMaksimenkau (changes requested):

AntonMaksimenkau (approved):

maxim12321 (approved):

PR #174, opened 2021-08-22
Don't allow to add listening topics without audios

Anlord241:


src/main/java/com/team4/testingsystem/services/impl/RestrictionsServiceImpl.java
```diff
@@ -202,4 +203,11 @@ public void checkNotArchivedContentFile(ContentFile contentFile) {
             throw new QuestionOrTopicEditingException("You can't edit archived topics");
         }
     }
+
+    @Override
+    public void checkListeningHasAudio(ContentFile contentFile) {
+        if (contentFile.getUrl() == null) {
```
AntonMaksimenkau, line 214: Let's also check that the file contains on s3

AntonMaksimenkau (changes requested):

src/main/java/com/team4/testingsystem/services/impl/RestrictionsServiceImpl.java
```diff
@@ -202,4 +207,19 @@ public void checkNotArchivedContentFile(ContentFile contentFile) {
...
+            throw new NoAudioException("You can't add listening topics without audios");
+        }
+    }
+
+    @Override
+    public void checkExistsOnS3(String fileName) {
+        Path filePath = Path.of(FileUtils.getTempDirectory() + "/testing-system/" + fileName);
```
maxim12321: I don't think this check is correct. Temp files aren't stored permanently, that's why they are temp :)

maxim12321 (changes requested):

src/main/java/com/team4/testingsystem/services/impl/RestrictionsServiceImpl.java
maxim12321: Amazon S3 SDK has this method:
`boolean doesObjectExist(String bucketName, String objectName)`
I think that's what you need

AntonMaksimenkau: We have several filestorages (at least local one and s3). And they have different logic for such checks. I think we should add isFileExist method to the interface and implement it on all the file storages. Then we should call the method here and throw the exception if file is not exist

AntonMaksimenkau (changes requested):

AntonMaksimenkau (approved):

src/main/java/com/team4/testingsystem/services/FilesService.java
```diff
@@ -11,4 +11,6 @@
     Resource load(String fileName) throws FileLoadingFailedException;
 
     void delete(String fileName) throws FileDeletingFailedException;
+
+    boolean isFileExist(String fileName);
```
maxim12321: Maybe `doesFileExist`?)

maxim12321 (approved):

PR #177, opened 2021-08-23
updated shuffle in question answers

Darina007:


maxim12321 (approved):

AntonMaksimenkau (changes requested):
Will postgress or JPA/Hibernate guarantee the order of the questions? In case it's not, we should make the sorting by ourselves. Otherwise, the pseudorandom will not guarantee the order as well.
P.s. Please, don't forget to check the solution actually work as expected :D

maxim12321 (commented):
BTW, maybe we can use, for example, hash of `question.id` and `test.id` as seed, not just `question.id`?

Darina007 (commented):
shuffling the answers works as it should, and I think that the order of the questions will not change, since after generation, when they receive a test, they are selected by the test id in a row without any mixing

AntonMaksimenkau:
> shuffling the answers works as it should, and I think that the order of the questions will not change, since after generation, when they receive a test, they are selected by the test id in a row without any mixing

We should definitely use the id for it. Alternatively, we could see all the correct answers as the first/second/... response

AntonMaksimenkau (approved):

AntonMaksimenkau:
> shuffling the answers works as it should, and I think that the order of the questions will not change, since after generation, when they receive a test, they are selected by the test id in a row without any mixing

The same test with the same test and questions id will have the same order. But it could shuffle the order of the answers for different tests.

PR #181, opened 2021-08-24
Don't allow coaches to verify verified tests

Anlord241:


Darina007 (approved):

src/main/java/com/team4/testingsystem/services/impl/RestrictionsServiceImpl.java
```diff
@@ -158,6 +158,13 @@ public void checkNotVerifiedForCoachDeassign(Test test) {
         }
     }
 
+    @Override
+    public void checkNotVerified(Test test) {
+        if (test.getStatus().equals(Status.VERIFIED)) {
```
maxim12321, line 163: Shouldn't we handle EXPIRED status?

maxim12321 (approved):

AntonMaksimenkau (approved):

PR #186, opened 2021-08-24
Add notifications for expired tests

maxim12321:


src/main/java/com/team4/testingsystem/converters/notifications/TestExpiredNotificationConverter.java
```diff
@@ -0,0 +1,21 @@
...
+    public NotificationDTO convertToDTO(Notification notification) {
+        return notificationBuilder(notification)
+                .level(notification.getTest().getLevel().getName())
+                .build();
+    }
+
+    @Override
+    public NotificationType converterType() {
+        return NotificationType.TEST_EXPIRED;
+    }
```
AntonMaksimenkau, line 20: I think, we can pretty easy make the same with level

src/main/java/com/team4/testingsystem/entities/Timer.java
```diff
@@ -25,9 +31,16 @@
...
+    private Status status;
+
+    @Column(name = "scheduled_time")
+    private Instant scheduledTime;
+
+    public Timer(Test test, Status status, Instant scheduledTime) {
         this.test = test;
+        this.status = status;
+        this.scheduledTime = scheduledTime;
     }
```
AntonMaksimenkau, line 45: Isn't it allargsconstructor?

src/main/java/com/team4/testingsystem/services/impl/TimerServiceImpl.java
```diff
@@ -0,0 +1,70 @@
...
+@Service
+public class TimerServiceImpl implements TimerService {
+    private final TimerRepository timerRepository;
+    private final ApplicationEventPublisher eventPublisher;
+
+    @Autowired
+    public TimerServiceImpl(TimerRepository timerRepository,
+                            ApplicationEventPublisher eventPublisher) {
+        this.timerRepository = timerRepository;
+        this.eventPublisher = eventPublisher;
```
AntonMaksimenkau: Allargsconstructor

Anlord241 (approved):

AntonMaksimenkau (approved):

src/main/java/com/team4/testingsystem/entities/Timer.java
maxim12321: Nope, field `id` isn't in the constructor

src/main/java/com/team4/testingsystem/services/impl/TimerServiceImpl.java
maxim12321: Maybe we should then use lombok for all services?

src/main/java/com/team4/testingsystem/converters/notifications/TestExpiredNotificationConverter.java
maxim12321: What means "the same"?
Is it about pushing `level` up to the abstract class?

AntonMaksimenkau: Yes

src/main/java/com/team4/testingsystem/services/impl/TimerServiceImpl.java
AntonMaksimenkau: We have a task for it

src/main/java/com/team4/testingsystem/converters/notifications/TestExpiredNotificationConverter.java
maxim12321: But `TestDeassignedNotificationConverter` doesn't set the level

PR #196, opened 2021-11-05
replace constructor with RequiredArgsContructor annotation in a few controllers; move dto and entity to model package; rename entities to entity

aretinsky:


AntonMaksimenkau:
Closed as the codestyle issues are not going to be fixed

